Issue #55, opened 2015-04-16
GenIso has embedded DNS nameservers not accessible


smalenfant:
Please take a look at the line setting up the network in the kickstart (GenIso.pm). There is a DNS name server not accessible outside of specific network.

Few enhancement possible

- DNS Server as GLOBAL/PROFILE
- kickstart_network_template to specify the network line to put in the ks.cfg. This one would allow different OS to be supported on the CDN (use case: migration)

PR #75, opened 2015-04-21
Change the way GenISO creates images. 

petrocc:
Instead of directly editing ks.src into ks.cfg, this change will write three files: disk.cfg, network.cfg and password.cfg. 

Additionally this change uses crypt to hash a provided password (Also sets a default if no password is provided. The default in the code is "Fred".) 

password.cfg is straight-up included into ks.cfg as an %include file. 

disk.cfg and network.cfg are intended to be processed by scripts.

### DIFF
--- a/traffic_ops/app/lib/UI/GenIso.pm
+++ b/traffic_ops/app/lib/UI/GenIso.pm
@@ -24,8 +24,21 @@ my $filebasedir = "/var/www/files";
 my $ksfiles_parm_name = "kickstart.files.location";
 my $ksfiles_configfile_name = "mkisofs";
 
+# This is the directory we put the configuration files in for kickstart &
+# scripts to process: 
+my $install_cfg = "ks_scripts"; 
+
 sub geniso {
 	my $self = shift;
+	print "here\n";
+	&navbarpage($self);
+	my %serverselect;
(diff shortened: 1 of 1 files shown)


dneuman64:
Line 33 has a print statement, can you remove?
Looks like lines 42 -51 are duplicate, can you remove?

elsloo:
Also, line 123 has a hardcoded name server IP that should not be there which was reported in issue #55.  Please resolve that issue somehow.

petrocc:
Fixed this and made the other suggested changes.

Regards,
Petro.
:wq

traffic_ops/app/lib/UI/GenIso.pm
```diff
@@ -75,99 +82,82 @@ sub iso_download {
...
+	} else {
+		$dir = $filebasedir . "/" . $osversion;
 	}
+	my $cfg_dir = "$dir/$install_cfg";
+	open (STUF,">$cfg_dir/state.out") or die "can't open state"; 
+	print STUF "Dir== $dir\n";
+	my $cmd = "mkisofs -joliet-long -input-charset utf-8 -b isolinux/isolinux.bin -c isolinux/boot.cat -no-emul-boot -boot-load-size 4 -boot-info-table -R -J -v -T $dir";
+	print STUF "$cmd\n";
+	close STUF;
+	# This constructs the network.cfg file that gets written in the $install_cfg directory
```
mtorluemke: Can we take these comments out, or make them generic?

PR #91, opened 2015-04-27
Two small fixes to deliveryservice api.

knutsel:


knutsel:
Added one more, because i don't know how pull requests work.

PR #100, opened 2015-04-28
This changes the fixed duration to a random duration.

petrocc:
Jan and I very, very briefly discussed this--when traffic_ops_ort runs they all run on the 15 and all delay 5 seconds, which means they all hit more-or-less at once. 

sleep_rand didn't have any rand in it, so I made it randier.


mtorluemke:
Thanks! Not sure what happened here...

PR #120, opened 2015-05-10
support for cacheurl.so as a remap plugin for dropping the querystrin…

### LINKED ISSUE #116: Need to support cacheurl plugin as a remap plugin, not a global.

knutsel:
…g out of cache key. Close #116


knutsel:
@markt: Just rename the config_file from cacheurl.config to cacheurl_remap.config, and it will start using remap. Lightly tested, so needs a review.

Issue #123, opened 2015-05-12
HTTP Request Routing Example (DS) doesn't match TR


smalenfant:
When you enable a delivery services via HTTP, the example shows a http://ccr.tld. Although after a while troubleshooting why it didn't work, found out that the Traffic Router is using "tr" instead of "ccr". 

I have looked in the database but couldn't nor find configuration for this, would be great to be able to change it to whatever we like.

UI Example
http://ccr.live.cdn.ltest.net

auto-zone config 
tr.live.cdn.ltest.net.	3600	IN	A x.x.x.x

PSUdaemon:
tr is correct and ccr in the example should be changed.

elsloo:
This is due to our transition to open source, and will be rectified under issue #142.  The UI says "ccr" and is not currently configurable or database driven.  The routing name can currently be configured by modifying /opt/traffic_router/conf/http.properties.  #142 will allow these to be user configurable.

knutsel:
See #142

PR #124, opened 2015-05-12
Change ccr to tr in example DS

PSUdaemon:
This is meant to address #123.


dneuman64:
I think this needs to be a more broad change in traffic_ops.  There are many places where ccr is referenced and we should address them all.

knutsel:
yea, I think there's a param that says what the prefix is, and we should use that?

Issue #137, opened 2015-05-14
Unable to export profile with no parameters


elsloo:
When attempting to export a profile with no parameters, the following error is displayed:

```
Traffic Ops fatal error occurred while processing your request. 
Error at line 387 (    my $text = JSON->new->utf8->encode($jdata);) 
hash- or arrayref expected (not a simple scalar, use allow_nonref to allow this) at /opt/traffic_ops/app/script/../lib/UI/Profile.pm line 387.
```

This stems from the query on line 372 that looks only at the profile parameter table in order to build the data structure that is marshalled to JSON on line 387.

We can fix this quickly by not allowing a parameterless profiles to be exported, or fix it thoroughly to support both cases (more time consuming and requires touching the import code as well).

knutsel:
why would you want to export a profile with no parameters?

elsloo:
Our Riak profiles have no parameters, which I was unaware of when I attempted to export the profile.  If we don't want to export a parameterless profile then we should provide a meaningful message instead of barfing.  I write a quick fix to do that if it's preferable over allowing the export to proceed without parameters.

PR #138, opened 2015-05-14
Do not allow users to export profiles with no parameters.  This close…

elsloo:
…s #137.


knutsel:
Why are there css changes with this commit? I don't understand...

elsloo:
Because I made the export profile button disabled when $param_count == 0.  It's better to see that it's disabled for a reason rather than simply removing buttons.  Doing that leads to a confusing user experience.

PR #163, opened 2015-05-23
Freeform cacheurl in deliveryservice view/table - @markt, pls review.

knutsel:
I'm keeping it empty by default for now, we can make a param later to set a default that takes the origin server out of the cache key.


knutsel:
Just realized, need to add docs, and also apply at the mids. Will add that to this PR tomorrow, so don't merge yet.

knutsel:
Added the mid and edge remap stuff. Should be good to go now.

PR #176, opened 2015-05-29
Issue173

mitchell852:
Global TO email settings should not be specific to the TO Portal - this will also require a change to /opt/traffic_ops/app/conf/cdn.conf to  look like:

```
to => {
		email_from => 'no-reply@to.domain.com', 
		base_url => 'https://to.domain.com'
    },
	portal => {
		email_from => 'no-reply@portal.domain.com', 
		base_url => 'https://portal.domain.com'
    },
```


dewrich:
Don't forget to update the docs/release_notes.txt and you'll need to follow this config change for every environment.

PR #180, opened 2015-06-01
Any map

knutsel:


traffic_ops/app/lib/UI/DeliveryService.pm
```diff
@@ -788,14 +782,13 @@ sub create {
 	return $self->redirect_to("/modify_error") if !&is_oper($self);
 	my $new_id = -1;
 
-	if ( !&is_oper($self) ) {
-		my $err .= "You do not have enough privileges to modify this.\n";
-		return $self->flash( message => $err );
-	}
+#	if ( !&is_oper($self) ) {
```
mtorluemke: Is this intentional?

knutsel:
yes. There's a return if you're not oper on the 3 lines above it. it's dead code.

PR #270, opened 2015-06-29
Fix semi colon

petrocc:
This should fix it.


dneuman64:
Thanks Petro but Jan fixed this already via #269

PR #296, opened 2015-07-09
New rpm generation for Traffic Ops

hbeatty:
I created a new rpm generation for Traffic Ops that uses a Vagrant environment.


mtorluemke:
Thanks for doing this work, and submitting the PR! 

Can you walk me through the decision you made on whether to have a single .spec and single build.sh between the two build types?

hbeatty:
It wasn't really about separating them. I was trying not to break what y'all already had working. Steve mentioned that Comcast is using Jenkins and automated builds. We don't have anything like that set up so I couldn't really test that I wasn't breaking anything.

mtorluemke:
Understood, thanks. We're excited about being able to use Vagrant with TO, but less excited about maintaining two versions of mostly the same code (if we make improvements to one version, will we remember to also make them to the other?). 

Should we merge as-is now, and see how much pain we endure, or should we try to combine the two methods now? 

I think our Jenkins build process is pretty simple -- just tries to replicate what can be done on the CLI manually.

hbeatty:
I really like the idea of combining them before the merge because I also do not like the thought of maintaining 2 different ways of generating the rpm. I also like the idea of an automated build process. I also like the idea of making it where someone that is not using Jenkins, ant, or any automated build process can generate an rpm.

I guess what the goal of the work I did was to give a way for someone to follow a set of very easy instructions and generate an rpm that they can then load on their system for testing their changes.

I actually tried to use the rpm build scripts that were already there. Here are some of the things I ran into:

* The build number is being generated by Jenkins and/or ant (from what I can tell).
* There are variables that I had no idea how they were being generated. (e.g. @VERSION@ in traffic_ops.spec)
* The install of ant is not straight forward. This was my main stopping point. I've never used ant before and I didn't have the time to figure out all the dependencies to get ant working properly.
* I have no idea what the XML file is for.
* There are some Comcast specific items in them (or seem to be).

We were also in a hurry because we wanted to test a change that we had made and thus the work that I did.

How would you like to proceed?

Have y'all thought about moving the building and rpm generation to something like [Travis](https://travis-ci.org/), [Fedora build service](http://koji.fedoraproject.org/koji/), or [Open Build Service](http://openbuildservice.org/)? I have no preference for one over the other and have not used any of them.

mtorluemke:
Thanks for that list -- give me a little time to discuss with folks here.

dangogh:
Just wanted to let you know that this is being actively looked at..   I now have a vagrant install as well as ant and the jenkins setup available,  so I'm going to see if these methods can be reasonably merged (and documented..).   It may take some time,  so stay tuned...

dangogh:
@hbeatty I think everything here is now covered by #517.  What do you think?

smalenfant:
We might want to see that Vagrantfile make it to the root of
traffic_control. This way one wouldn't have to "git clone" twice. Comments?

hbeatty:
Where does GIT_URL get defined? I tried following the Readme.md and I get the following error:
```
ERROR: GIT_URL is null. Check the project config and insure that git SCM is configured.
```

dangogh:
I'll check on both of these...  thanks..

dangogh:
@hbeatty I think I've eliminated GIT_URL completely..   Is there something left over in your directory that's causing that?

hbeatty:
I was using dangogh/noant to do my testing. Is this not the latest?

dangogh:
yes -- that's the latest, and I just did a clean clone followed by 'git checkout noant' to make doubly sure..  There are no files in there with GIT_URL.

hbeatty:
I'm sorry. I tried what you said and I still can't make it work for me. I must be missing something.

I've tried multiple ways of doing it. If you would like I could try to fix it. and then do a pull request to your repo.

Issue #42, opened 2015-04-14
download_web_deps is not executed by the postinstall script.


knutsel:
Currently have the docs saying to do it manually, but the postinstall script should do the right thing.

PR #342, opened 2015-07-24
move PERL5LIB setting further up for all cmds

dangogh:
PERL5LIB not set early enough, so download_web_deps fails silently with missing dependencies


knutsel:
This closes #42 , right? Thank you!

dangogh:
I have another change that makes the failure visible..   coming shortly..

PR #343, opened 2015-07-24
make exec failures visible using IPC::Cmd instead of fork/exec

dangogh:
download_web_deps error was masked by this.   Error will look something like this:

Running [/sbin/service traffic_ops start]...
ERROR: /sbin/service traffic_ops start failed


knutsel:
Do we need to add that new IPC::Cmd dependency to cpanfile?

dangogh:
nope..   it was already there..

PR #373, opened 2015-08-03
Fix/active servers

jheitz200:
Most of the changes are based on feedback from go vet, golint, and go fmt.


dneuman64:
way to make my code look better

PR #374, opened 2015-08-03
Feature/move os dirs

dewrich:


dewrich:
This PR will require a local update to your Perl dependencies:

$ cd traffic_control/traffic_ops/app (where your cpanfile lives)
$ carton

PR #389, opened 2015-08-05
add dns_bypass_cname to deliveryservice

dangogh:


elsloo:
45 characters seems a little short for the column size given the length of some domain names we would use with this feature. Maybe something like 128 characters instead?

dangogh:
good point -- is length expected to be different from http_bypass_fqdn?

mtorluemke:
nope -- using the same 255 should suffice.

dangogh:
changed to 255

PR #413, opened 2015-08-10
this puts the redis main graph back.

knutsel:
This PR puts the main Redis graph back as **Health->Graph View (old)**, and makes the Influx/Grafana view available as **Health->Graph View (new)**. All the sub links are still going to Grafana/Influx, and the Redis view only supports 24 hours. 

Not sure if this PR will make the transition to Grafana/Influx easier and less risky. 

@dneuman64,  @mtorluemke : your call. Don't merge this PR if you think it'll make things more messy without adding value.


dneuman64:
I can live with it.  Will give @mtorluemke a little while to think about it.  If he approves I can merge tomorrow.

PR #420, opened 2015-08-12
add syslog and combined MTU to ToPingCheck.pl

hbeatty:
- added some example cron entries
- incremented version
- removed $hostn var
- set syslog to LOG_INFO
- removed $match var
- added $check_lng_name var - configured from $jconf via command line
- added some trim functions

### DIFF
--- a/traffic_ops/app/bin/checks/ToPingCheck.pl
+++ b/traffic_ops/app/bin/checks/ToPingCheck.pl
@@ -15,8 +15,17 @@
 # limitations under the License.
 #
 
-# Plugin for the "ping" check.
+# Plugin for the "ping" and "MTU" check.
 #
+# example cron entry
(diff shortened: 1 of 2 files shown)


traffic_ops/app/bin/checks/ToPingCheck.pl
```diff
@@ -28,10 +37,9 @@
 use Log::Log4perl qw(:easy);
 use JSON;
 use Extensions::Helper;
+use Sys::Syslog qw(:standard :macros);
```
mtorluemke: @hbeatty: Can you add Sys::Syslog to the cpanfile?

hbeatty:
@mtorluemke I added Sys::Syslog to the cpanfile. Thanks, Hank

PR #422, opened 2015-08-12
added syslog and IPv6 checking to ToDSCPCheck.pl

hbeatty:
1. traffic_ops/app/bin/checks/IPv6.pm - new file, cannot be installed
by CPAN.
2. traffic_ops/app/bin/checks/ToDSCPCheck.pl
  * added IPv6 checking
  * incremented version
  * removed $hostn
  * changed the way autoflush is handled
  * added $cms_int var now required on the command line in jconf
  * removed $match because it was not being used
  * added $chck_lng_nm for syslog messages
  * I was going to put in to have it fail if the ‘checkPath’ wasn’t
defined but, opted not to. Left the code in but commented.
  * changed curl to use the -H option and the url now uses the IP
address
  * changed get_dscp to also return -1 so that if we can’t connect to
the server we will know.
  * added trim functions.

### DIFF
--- a/traffic_ops/app/bin/checks/IPv6.pm
+++ b/traffic_ops/app/bin/checks/IPv6.pm
@@ -1,0 +1,378 @@
+#
+# IPv6.pm
+# NetPacket::IPv6
+#
+# Decode Internet Protocol v6 packet header.
+#
+# References:
+# RFC2460 - IPv6 Specification
+#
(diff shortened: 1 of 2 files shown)


hbeatty:
I should have mentioned that the IPv6.pm file needs to go in /opt/traffic_ops/lib/perl5/NetPacket/. There isn't a lib dir off traffic_ops until the install so I added to this dir.

Issue #349, opened 2015-07-27
traffic_ops 1.1.4 postinstall cannot install a perl module


lwickline:
post install fails with 
Installing modules using /opt/traffic_ops/app/cpanfile
! Installing File::Copy::Recursive failed. See /root/.cpanm/work/1438026616.16189/build.log for details. Retry with --force to force install it.
! Installing the dependencies failed: Module 'File::Copy::Recursive' is not installed
! Bailing out the installation for /opt/traffic_ops/app/.
Installing modules failed

the error in the log is:
Building File-Copy-Recursive-0.38
make: chmod: Command not found
make: *** [blib/lib/File/Copy/.exists] Error 127
-> FAIL Installing File::Copy::Recursive failed. See /root/.cpanm/work/1438026616.16189/build.log for details. Retry with --force to force install it.
-> FAIL Installing the dependencies failed: Module 'File::Copy::Recursive' is not installed
-> FAIL Bailing out the installation for /opt/traffic_ops/app/.

PR #424, opened 2015-08-12
use system() to avoid IPC::Cmd dependency;   fix PATH and PERL5LIB for module makefiles

dangogh:


dneuman64:
this closes #349

dangogh:
yes -- forgot to add that..  thanks..

PR #438, opened 2015-08-18
Fix parent.config when mulitiple msites are configured.

### LINKED ISSUE #437: parent.config has all mutlisite origins in all destinations
when multiple different mutlisite configs are present.

knutsel:
This fixes #437


traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -998,7 +1015,7 @@ sub parent_dot_config {
 		my %done = ();
 
 		foreach my $remap ( @{ $data->{dslist} } ) {
-			if ( $remap->{type} eq "HTTP_NO_CACHE" || $remap->{type} eq "HTTP_LIVE" || $remap->{type} eq "DNS_LIVE" ) {
```
mtorluemke: Is this intentional?

knutsel: No! Good catch. I was taking this from 113... <blush> Will fix.

PR #441, opened 2015-08-18
Issue 415

dangogh:
increased inactivity timeout to be more in line with external sources (e.g. spdb)


dangogh:
brought up-to-date with current base branch

PR #445, opened 2015-08-18
Fix parent selection in multi multi stite config. 

knutsel:
Cherry picked from master

git cherry-pick ca04d3f
git cherry-pick 72780fc
git cherry-pick a184fe7


PSUdaemon:
For future reference `git cherry-pick -x` will give you more context in your commit messages so you don't need to leave a comment as to where you cherry picked them from.

PR #467, opened 2015-08-25
Add syslog messages to ToORTCheck.pl

hbeatty:
* added example cron entries
* removed $username = “jvando001”
* incremented version
* removed $hostn
* added -h option
* added help()
* re-ordered beginning so that script would error quicker if command line vars
were incorrect
* added $chck_lng_nm for syslog messages
* added $to_user and $to_pass for ORT script on remoter server
* removed $match and $select
* added code to get Traffic Ops URL from TO API
* changed the ssh command to include the $to_url, $to_user, and $to_pass
* added 2>&1 to ssh command so that I could check if ssh was unable to
connect to server (added -1 for the ORT column in TO)
* Will now output results to /var/www/html/\<host_id\>/results (this file
gets overwritten each time the script is run)
* ERRORS, FATALS, and WARNS are calculated for syslog messages
* only ERRORS are reported for the TO column (this is unchanged from
how the script worked before)
* This script should be run using a user that has an ssh key on the
edges and mids in the authorized_keys file. That user will also need
sudoers access on the edges and mids.

### DIFF
--- a/traffic_ops/app/bin/checks/ToORTCheck.pl
+++ b/traffic_ops/app/bin/checks/ToORTCheck.pl
@@ -16,40 +16,37 @@
 #
 # ORT check extension. Checks how many errors there are running the "ort" script on the cache
 #
+# example cron entry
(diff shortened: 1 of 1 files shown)


knutsel:
Hi Hank,

Thanks!

I've been considering to move this cron to the caches themselves and have them post the result to TO from there... That way we get rid of all this ssh stuff. There already is an ort cron on the caches to syncds.... 

Since you're in there anyway... Want to tackle that?

hbeatty:
I'd be happy to.

knutsel:
Sweet! Should we merge this first or wait for that?

hbeatty:
If you don't mind I would like to go ahead and merge this.

I'll create a new branch for the new changes.

Issue #317, opened 2015-07-17
Add display_name column to deliveryservice table


mitchell852:
type = varchar(48)
null = no

initially seed the column with the value from xml_id

make sure to add this new field to the TO delivery service form.

mare sure to update the payload of the following apis as well:

/api/1.1/deliveryservices/:id
/api/1.1/deliveryservices

mitchell852:
i see the column as

| Field                  | Type          | Null | Key | Default           | Extra                       |
| display_name    | varchar(1024) | YES  |     | NULL              |                             |

can it be changed to:

varchar(48) <-- the traffic portal handles long values but people should keep it short to avoid to much truncation
null = no (the traffic portal depends on a display name now)

mitchell852:
and actuallly, while you're at it, i would recommend increasing the long_desc* fields to varchar(1024)

knutsel:
@dangogh - can you take a look??

mtorluemke:
@mitchell852: okay to close?

mitchell852:
yep, looks good to me

PR #470, opened 2015-08-25
TO: deliveryservice/display_name shortened from 1024 to 48

dangogh:
Issue #317 : display_name shortened from 1024 to 48


dangogh:
updated to not allow null in form, save on create.   Should be complete now.

Issue #491, opened 2015-09-02
In the traffic ops api, purge / invalidate content requests are not being limited to the delivery services assigned to current user 


mitchell852:
POST /api/1.1/user/current/jobs

{dsId: "9999", dsXmlId: "deliveryservice-name", regex: "/foo.html", ttl: 48, startTime: "2015-09-02 11:02:12"}

^^ this will create a "purge" job for dsId=9999 regardless of whether the currently authenticated user has access to the 9999 delivery service or not.

the correct result should be a 403 forbidden if delivery service 9999 is not assigned to the user....unless user is admin

mitchell852:
maybe operations role can skip the "is delivery service assigned" check as well...

mitchell852:
also i updated the documentation related to purge - https://github.com/mitchell852/traffic_control/commit/a2a90a64923b64bcd40e282088c15d4380a5bb20

mitchell852:
looks good

mitchell852:
actually, pull request #496 didn't make it in to the 1.1.5 release so it will have to go into 1.1.6 (but the changes that went in still satisfy the requirement of preventing the ability to create a purge job against an unassigned delivery service)

PR #495, opened 2015-09-03
add check for user authorized to create purge job

dangogh:
For issue #491 -- only admin, operations, or assigned user can submit purge job on a delivery service.


dewrich:
This fixes https://github.com/Comcast/traffic_control/issues/491

PR #497, opened 2015-09-04
modify method of generating rpm

dangogh:
Based on hbeatty's pull request, but made build_rpm.sh better able to deal with non-Jenkins environment;  Vagrantfile included to allow for use of vagrant in generating the rpm.

This is in need of review before merging.


hbeatty:
@dangogh I like it. I haven't had a chance to test but, I will. What do you think of using the # of commits and the commit sha for the build? I noticed you left a TODO in there about getting the build from the repo instead of just incrementing what was in the file.
```
   COMMITS=$(git shortlog | grep -E '^[ ]+\w+' | wc -l)
   SHA=$(git rev-parse --short=8 HEAD)
   BUILD_NUMBER="$COMMITS.$SHA"
```

dangogh:
@hbeatty I do like the method you used.  It provides both a way to know how a build falls within the sequence of builds as well as an identifier that allows grabbing the right version from the repo.  I've had a bit of trouble getting the two build methods meshed, so I decided to leave that out while testing this.  I think it's important enough to warrant a separate issue.

PR #499, opened 2015-09-09
Use ATS style key value formatting for http events in the access log

trevorackerman:
Rally - US265
Start making TR access logging more in line with ATS
Should get two log entries per request to TR at port 8081, the initial request and then the final response
No intermediate routing decisions logged yet as part of access


elsloo:
Can you build this and take a look at the PMD warnings for variables that could be final that aren't marked as such?  There are several areas that I believe will generate warnings due to, for example, method arguments not being marked as final.

trevorackerman:
PMD is happier now

limited:
Why are two separate access log entries needed for a single client request? 

Couldn't all information be held in a single access log entry that is logged when the TR is done servicing the request?

trevorackerman:
One logging event per single client HTTP request.
Added DNS query access events to log

PR #500, opened 2015-09-09
check for go and ant in path

dangogh:
if either go or ant not available in PATH or go compile fails, exit with a reasonable message and non-zero status


dangogh:
this is for TO rpm generation..

Issue #512, opened 2015-09-11
TO: __RETURN__ should add newline for cacheurl_(.*).config


mtorluemke:
__RETURN__ adds a newline for header rewrite, the same should work for cacheurl config files.

mtorluemke:
__RETURN__ has become a convention to insert newlines into remap config files -- header rewrite is probably the best example for this.

PR #548, opened 2015-09-22
substitute \n for __RETURN__ in cacheurl

dangogh:
#512 -- does substitution of \n for __RETURN__ in cacheurl.   Still requires some testing


dneuman64:
this closes #512

PR #549, opened 2015-09-23
Add result type details code as a suffix to the rtype field in access log

trevorackerman:
e.g. rtype=MISS.DS_NO_BYPASS


trevorackerman:
re-doing on a different branch

PR #552, opened 2015-09-23
mvn clean verify will now fail if new code creates new PMD warnings

trevorackerman:
suppress existing pmd warnings
tweak pom file so that pmd warnings fails build


trevorackerman:
This pull request is now stale

Issue #131, opened 2015-05-13
diff tool with parameters missing entries


G4m30v3r:
when doing a compare of profiles, for example cache url.so states it belongs to profile x, but does not show up when selecting profile x.

![](https://files.slack.com/files-pri/T024VU91V-F04S7BM57/pasted_image_at_2015_05_13_03_58_pm.png)

![](https://files.slack.com/files-pri/T024VU91V-F04S7BWNX/pasted_image_at_2015_05_13_03_58_pm.png)

G4m30v3r:
https://github.com/Comcast/traffic_control/issues/134

mtorluemke:
@PSUdaemon has more info to add.

PSUdaemon:
My proposed solution would be to do this in the database, either by DBI or custom query. If you do a full outer join on two aliased versions of the profile parameters table by key where value is different or NULL then you should get the exact data you want and can just print it in Perl rather than having to do some crazy data manipulation in Perl that is hard to prove correct. This will also make doing it in another language easier. So you don't have to prove it correct in Perl/Python/Go/etc.

PSUdaemon:
More follow up. MySQL does not support full outer joins. There is some workaround that makes the query kinda messy but doable.

PR #559, opened 2015-09-29
Fix Profile Compare

PSUdaemon:
This takes into account the case where just config files differ. It also refactors the code to make it simpler and has some cleanup of unused code.

This addresses #131.


PSUdaemon:
Param name can have dups.

PR #572, opened 2015-10-01
Combine ts_daily_summary into write_traffic_stats

PSUdaemon:
This merges ts_daily_summary into write_traffic_stats so now there is one process to run instead of two.

Also, this updates write_traffic_stats to handle the new cdnName changes.


PSUdaemon:
After offline discussion with @dneuman64 I made the entire package more generalized as "traffic_stats" and fixed up a few things like sending stats before we shutdown and making config reloadable.

traffic_stats/traffic_stats.go
```diff
@@ -122,32 +154,211 @@ func main() {
...
+					continue
+				}
+				max = FloatMax(max, kbps)
+				duration := sampleTime.Unix() - prevtime.Unix()
+				bytesServed += float64(duration) * kbps / 8
+				prevtime = sampleTime
+			}
+			maxGbps := max / 1000000
+			bytesServedTb := bytesServed / 1000000000
+			log.Debugf("max gbps for cdn %v = %v", cdn, maxGbps)
```
dneuman64: Can we make this log.Infof?

traffic_stats/traffic_stats.go
```diff
@@ -122,32 +154,211 @@ func main() {
...
+			log.Debugf("Tbytes served for cdn %v = %v", cdn, bytesServedTb)
```
dneuman64: this one log.Infof too

traffic_stats/traffic_stats.go
```diff
@@ -248,7 +464,7 @@ func calcMetrics(cdnName string, url string, cacheGroupMap map[string]string, co
 	} else if strings.Contains(url, "DsStats") {
 		err = calcDsValues(rascalData, cdnName, sampleTime, config)
 	} else {
-		log.Info("Don't know what to do with ", url)
+		log.Debug("Don't know what to do with ", url)
```
dneuman64: lets make this a warn

PR #592, opened 2015-10-05
Grafana dashboards

smalenfant:
To support grafana 2.1.3 (works with older version as well)


dneuman64:
Are these 3 meant to replace traffic_ops_scripted.js?  Are they in addition too?

smalenfant:
They are meant to replace traffic_ops_scripted.js. I kept this one around because it still works with griffin 2.0.2.

dneuman64:
ok, i will merge and give them a look.  We will need to get the documentation updated to reflect the new scripts.

PR #595, opened 2015-10-05
Changing a few graphs and links

smalenfant:
- Updated the Server graph URL with Traffic Stats (griffin)
- Added INFLUXDB and TRAFFIC_STATS URL


smalenfant:
Be careful... Not too sure why so many changes.. probably because of the merging of my fork (hopefully)

PR #596, opened 2015-10-05
Server graph URL

smalenfant:
Changing the server graph from before to Grafana.
- Added TRAFFIC_STATS URL
- Added INFLUXDB URL


dneuman64:
I have concerns about the TRAFFIC_STATS and INFLUXDB types you added.  I like the idea but by default traffic stats is not listening on port 443 (grafana is) also do we really want to hard code the influxdb server to http and port 8083?  I think maybe it's best to just leave those out for now?

smalenfant:
These can be removed for now. They work for us with the default configuration which we followed with Traffic Stats. Although I could possibly add some TRAFFIC_STATS url template for Grafana and influxdb in the profile parameters.

PR #601, opened 2015-10-06
Traffic Stats Spec file and documentation

dneuman64:
Updated the Spec file to copy the Grafana custom dashboard files into place as part of the RPM install process.
Updated the Traffic Stats admin docs to reflect the change.


PSUdaemon:
You need to add an entry to the %files section for those files you copied as well.

dneuman64:
good catch, @PSUdaemon

smalenfant:
Should we had a dependency on grafana rpm to be installed?

PSUdaemon:
@smalenfant I think that is a good idea, but in that case i think we should separate it out into two RPM's.

dneuman64:
Adding it as a dependency is a good idea but we do not require grafana and traffic stats to be installed on the same server, so I do not think we should make it a dependency.  I put a comment in the docs that said if you put traffic stats and grafana on the same server then the files will be copied there for you, otherwise you need to do it yourself.

PSUdaemon:
We can think about the 2 RPM's in the future if need be.

PR #609, opened 2015-10-06
Postinstall prompts

dangogh:
not ready to pull in, but ready for someone else to try out..   I've redone the prompting in postinstall to be more consistent and to get initial values from conf files.   This should allow us to get closer to detecting already set up installations and bypass many (all?) of the prompts.


dangogh:
further updates coming with new pull request -- closing this one out..

PR #610, opened 2015-10-07
adding 'ALL' CDN

mtorluemke:


PSUdaemon:
Should we make the column `not null` as well?

mtorluemke:
thanks. should be good now.

PR #617, opened 2015-10-08
refactor

jheitz200:
refactored gen_traffic_router_config to limit the number of db queries


dewrich:
Code has drifted too far, won't merge

PR #624, opened 2015-10-09
Feature/federation api

jheitz200:
adding federation user, and federation types to seeds.sql.
adding external get api for federations.


jheitz200:
sorry put it in the wrong place !

PR #625, opened 2015-10-09
cdn.conf validation checks for secrets instead of shared_secret

dangogh:
Doesn't affect function of secrets,  only warning message for cdn.conf.


dangogh:
Fixes #631

PR #641, opened 2015-10-13
Changing "Drop at Edge" Query String handling

smalenfant:
Not quite ready to merge. It compiles and tested OK.
- Adding pparam=no-query-string on dscp remap (seems always present)
- Removing drop_qstring generation altogether as it doesn't work without header rewrite


knutsel:
@smalenfant - sorry this has taken so long... Just seems there's a lot of details here that could go wrong. You use this in prod, yes? You edit the official release to get this code in? Also, do you use it in conjunction with signed urls?

smalenfant:
Thanks for the feedback... Not running in production, but was running in the lab. I'll need to run a bunch of test cases. I'll keep updating the Pull Request until it's ready.

smalenfant:
Closing this pull-request for time being. Looks like I mixed some regex_remap with header_rewrite. Still looking for a solution.

Issue #196, opened 2015-06-03
Traffic Router sends NXDOMAIN for http.routing.name when no caches are assigned to the delivery service


elsloo:
Traffic Router sends NXDOMAIN for http.routing.name when no caches are assigned to the delivery service.  We should not send NXDOMAIN, we should send the Traffic Router IPs, then 503 when the client reaches the HTTP side of Traffic Router.

PR #647, opened 2015-10-14
Added domains array to deliveryServices section of CRConfig

dneuman64:
traffic ops now returns a deliveryServices->domains section in the CRConfig


dneuman64:
This is the first part of a fix for #196.  Traffic Router changes are next...

PR #650, opened 2015-10-15
Refactored influxConnect duplicate code

robert-butts:
This removes the duplicate connect code in influxConnect() by putting it in a local function.

I just noticed influxConnect has been refactored in master, so this might not be worth it.


PSUdaemon:
Where did you branch this from? This looks like code that isn't from master. Take a look at commit 5a8273d.

PSUdaemon:
In fact, this even refers to ts_daily_summary.go which no longer exists. Both scripts were merged into one file.

PSUdaemon:
Sorry, just re-read the second half of your comment! :-p Yeah, I am going to close this.

PR #660, opened 2015-10-16
update traffic_ops install doc

Vuuo:
Feels like the doc is out-dated, some are misleading.

traffic_ops package is not available in any default yum repositories that come with CentOS.
Step 3 has been implemented in the installer ever since the init commit.


PSUdaemon:
@Vuuo, please sign the CCLA before we can accept this PR.

More info here:

https://github.com/Comcast/traffic_control/blob/master/CONTRIBUTING.md

Thanks.

PSUdaemon:
Received CCLA. Thanks.

PR #664, opened 2015-10-16
Fix installation profile parameters

### LINKED ISSUE #661: Profile parameter don't match parameters, failing clean installation
This commit <https://github.com/Comcast/traffic_control/commit/39f4a0d4546e57ebd8a3dcf0e00730407ae34176> broke the install past 1.1.3. We have reverted the file and the install goes properly (after copying some Go binaries) 

LogFormat probably still need to be fixed afterward.

In addition, all the "Go" programs that came with the build are not present after that build.

smalenfant:
Fixes #661 against 1.1.6. This will give a good idea how to install 1.1.6. Another pull request will come in against 1.2.x soon.


smalenfant:
Not exactly what I wanted to do...

Issue #671, opened 2015-10-19
Delivery Service defaults to can "ALL"


smalenfant:
In that condition, you can't assign any servers.

jheitz200:
We have removed "ALL" from all fresh installs of Traffic Ops and that will now be an internal value.

Issue #670, opened 2015-10-19
Traffic Ops 1.2.1 - Cannot configured Delivery Service correctly


smalenfant:
Having some problem configuring Delivery Services in 1.2.1. Please see screenshot using Chrome and/or Safari.

<img width="512" alt="ds_screen_issue" src="https://cloud.githubusercontent.com/assets/3803592/10590601/9ce6cf48-767f-11e5-9fbc-c762cbc95085.png">

jheitz200:
@smalenfant are you still seeing an issue with this?

smalenfant:
@jheitz200 No, this can be closed. Although there is still a bug in there if you select ANY_MAP and go back type DNS for example, the fields don't get reset properly on the page.

jheitz200:
ok, there is another issue for ANY_MAP https://github.com/Comcast/traffic_control/issues/740

I am going to close this issue, please re-open it if you see problems again.
Thanks
Jackie

PR #672, opened 2015-10-19
Updates for 1.2.1

dneuman64:
Removed the ALL CDN migrations
Changed the sort of DeliveryService types to Description so that the ANY_MAP CDN is not first.


dneuman64:
this addresses #670 but a better fix should be put in place

dneuman64:
this closes #671 because the ALL CDN should not be there on a clean install any more

PR #684, opened 2015-10-22
Appropriately reduce traffic router application logging

trevorackerman:


trevorackerman:
Re-did this in another PR.

PR #688, opened 2015-10-22
Just do a one liner log entry instead of stack traces for malformed dns requests

trevorackerman:


traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/dns/protocol/TCP.java
```diff
@@ -99,6 +100,8 @@ public void run() {
                 final byte[] response = query(client, request);
                 os.writeShort(response.length);
                 os.write(response);
+            } catch (final WireParseException e) {
+                LOGGER.info(e.getMessage());
```
elsloo: Can we change this to warn or error?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/dns/protocol/UDP.java
```diff
@@ -105,8 +106,10 @@ public void run() {
                 final byte[] response = query(client, request);
 
                 final DatagramPacket outPacket = new DatagramPacket(response, response.length,
-                        packet.getSocketAddress());
+                    packet.getSocketAddress());
                 getDatagramSocket().send(outPacket);
+            } catch (WireParseException e) {
+                LOGGER.info(e.getMessage());
```
elsloo: Can we change this to warn or error?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/dns/protocol/TCP.java
trevorackerman: AbstractProtocol.query is already adding it to the access log:
1446066071.653 qtype=DNS chi=10.169.98.91 ttms=0 xn=- fqdn=- type=- class=- ttl=- rcode=- rtype=- rloc="-" rdtl=- rerr="Bad Request:WireParseException:end of input" ans="-"

Feels like that's enough but I'm willing to bump it up in the application log too, it'll just be duplicated is all.

elsloo: I don't follow. The exception is logged at INFO which means it will be in both the access log and application log anyway; if you wanted to avoid duplication, wouldn't you want to swallow the exception or log it at a level other than INFO? In any case, INFO is not correct for an error?

trevorackerman:
Closing this since there are merge conflicts. Easier to just re-do in a fresh branch.

PR #697, opened 2015-10-22
only allow DS to be created in a CDN if it has caches in it

mtorluemke:


PSUdaemon:
TBH, I am not sure I agree with the concept of this patch. Seems like a hack to address shortcomings of another hack. I understand the DS can't be active on a CDN with no caches, but this forces an order of operations that seems arbitrary. One could argue there is no point in adding caches to a CDN with no DS's. It's circular logic and you are just validating one side of it.

mtorluemke:
Yes, good point. Should we then decorate each CDN (new column in CDN table) to designate non-cache CDNs? Or should we just rely on the operator to know not to put delivery services in the wrong CDN?

PSUdaemon:
I strongly lead toward the latter. We do have protection to not allow assigning DS's from one CDN to caches of another CDN, correct?

mtorluemke:
Yes, that exists.

PSUdaemon:
Yeah, I just think that this wouldn't be intuitive for people without a "non-cache" CDN.

I could support the compromise of a warning if you add a DS to a CDN with no caches, but with the caveat that you also need a similar warning when adding caches to a CDN with no DS's. Perhaps even with the flow leading you into selecting caches to add a DS too in the non-error case (or DS's to add to a new cache).

mitchell852:
hate to comment after this is merged / closed but /api/1.2/cdns.json?types=EDGE,MID gives the impression we have cdns of type edge and mid. not sure if we're going to document this query parameter but if so, it's going to make it tougher to explain.

 /api/1.2/cdns.json?withCaches=true would be easier to explain.

mitchell852:
oh nm, this was kiboshed, huh? not merged. my bad.

Issue #704, opened 2015-10-23
Queue updates doesn't work in 1.2.1.


knutsel:
Select a CDN, select a CG (or set to all), click queue updates. No errors, nothing bad in logs, but the clocks don't appear in the server checks page, and the upd_pending isn't set.

PR #705, opened 2015-10-26
Fixes 704 - Queue updates

jheitz200:
Updating queue update route so that cachegroup names get passed in correctly.
Removing unused code `traffic_ops/app/lib/UI/Update.pm`


PSUdaemon:
This fixes #704.

PR #707, opened 2015-10-26
Fixes 704 - Queue updates

jheitz200:
Updating queue update route so that cachegroup names get passed in correctly.
Removing unused code `traffic_ops/app/lib/UI/Update.pm`


PSUdaemon:
I will merge this but I am going to revert the perl-tidy commit because it doesn't belong in the release branch.

PR #721, opened 2015-10-27
build all sub-projects within traffic_control

dangogh:
**NOT** ready for pulling in,  but I'd like feedback..   still requires more testing.  Top script is ./build/build.sh which invokes build/build_rpm.sh in each of the sub-projects.

traffic_monitor and traffic_router don't inherit version number (VERSION.txt) and build number isn't consistent, yet.


build/build.sh
```diff
@@ -0,0 +1,45 @@
...
+
+for p in $projects; do
+	echo "-----  Building $p ..."
+	case $p in
+		traffic_ops)     (cd traffic_ops/build && ./build_rpm.sh) ;;
+		traffic_ops_ort) (cd traffic_ops/build && ./build_ort_rpm.sh) ;;
+		traffic_monitor) (cd traffic_monitor/build  && ./build_rpm.sh) ;;
+		traffic_router)  (cd traffic_router/build && ./build_rpm.sh) ;;
+		traffic_stats)   (cd traffic_stats/build && ./build_rpm.sh) ;;
+		*) echo "No project named $p"; exit 1;;
```
PSUdaemon: Is it possible to merge the ort build into the traffic_ops build and have more generic code here rather than listing everything out in a case statement?

VERSION.txt
```diff
@@ -0,0 +1 @@
+1.2.1
```
PSUdaemon: To be consistent I think we should remove the .txt extension.

dangogh: I wasn't sure if we wanted version to be in its own file or in a more generic json or cfg..  what are we being consistent with?

traffic_monitor/build/build_rpm.sh
```diff
@@ -0,0 +1,113 @@
...
+# limitations under the License.
+#
+
+#----------------------------------------
+function buildRpm () {
+	echo "Building the rpm."
+
+	local version="-Dtraffic_control.version=$TC_VERSION"
+	local targetdir="-Dproject.build.directory=$(pwd)"
+	export GIT_REV_COUNT=$(git rev-list HEAD | wc -l)
```
PSUdaemon: Someone showed me `git rev-list --count HEAD` recently and that seems more succinct to me.

build/build.sh
dangogh: yes -- I was leaning toward more generic, but it started out with each using a different method for each.   I could eliminate the list of projects there altogether and just do all subdirs with build/build_rpm.sh.

dangogh:
yes on functions.sh -- as I said,  I know that part was incomplete, but good enough to have you all have a look.     and --count isn't available on git installed by yum on Centos 6.[56] -- that's why I used that method instead.

VERSION.txt
PSUdaemon: Other top level files. They are all caps and no extension or have a .md extension. NOTICE and LICENSE specifically are text files but don't have the .txt extension.

PSUdaemon:
Ah, that makes sense RE the git versions on CentOS.

dangogh:
really good feedback..   thanx..

traffic_stats/build/traffic_stats.spec
```diff
@@ -20,24 +38,43 @@ Installs traffic_stats which performs the follwing functions:
...
+  cp -r "$TC_DIR"/traffic_ops/client/* . && \
+  go get -v \
+) || { echo "Could not build go program at $(pwd): $!"; exit 1; }
+
+godir=src/github.com/comcast/traffic_control/traffic_stats
+( mkdir -p "$godir" && \
+  cd "$godir" && \
+  cp -r "$TC_DIR"/traffic_stats/* . && \
+  go get -v \
+) || { echo "Could not build go program at $(pwd): $!"; exit 1; }
```
PSUdaemon: Big :+1: on this.

So, `rpmbuild` knows how to do the rest of the build here? ie, `go build`?

dangogh: no -- "go get" does it all -- it gets and installs the dependencies,  then builds and installs the exec as well.

traffic_stats/build/traffic_stats.spec
```diff
@@ -20,24 +38,43 @@ Installs traffic_stats which performs the follwing functions:
 %setup
 
 %build
+export GOPATH=$(pwd)
+# Create build area with proper gopath structure
+mkdir -p src pkg bin || { echo "Could not create directories in $(pwd): $!"; exit 1; }
```
PSUdaemon: FWIW, once you have the GOPATH set, the `go` tool will make all these directories for you. Even if you did `go get github.com/Comcast/traffic_control`, it will create the `src` dir. I know you copy that over since we are building off a specific branch/tag though.

But that also brings up the point that we need to figure out a way to do proper dependency tracking too. For example, `traffic_stats` needs a specific version of the InfluxDB client. They have changed it in master and broken compatibility. Also, just for our own `traffic_ops` client, we need to align versions since the `CDN_name` change. Right now we manage that all by hand.

PSUdaemon:
Overall this looks very good. Definitely moving in the right direction.

traffic_stats/build/traffic_stats.spec
PSUdaemon: Ah, I did not realize that would make the project in the CWD as well. Cool.

traffic_stats/build/traffic_stats.spec
dangogh: did not know that about dir creation -- thanks..  and is there a known "good" way to handle lib versions?

traffic_stats/build/traffic_stats.spec
dangogh: yes -- you can run build_rpm.sh directly in any of them,  or give top-level build.sh a list of subdirs to build.

traffic_stats/build/traffic_stats.spec
PSUdaemon: There are multiple ways to do it, each with their own pros and cons. I think we need to do more research here.

[Here is a wiki](https://github.com/golang/go/wiki/PackageManagementTools) that lists out most of the options available.

dangogh: agreed afa more research..  I'll finish this and we'll address that separately.

dangogh:
will open new pull request after updating..

Issue #507, opened 2015-09-11
TO: rpm build exit status not consistent


dangogh:
build_rpm.sh needs to check for error status on each command.  Some are currently checked, but many are not,  and missing a bad command can show "good" status in Jenkins on a failed build.

mtorluemke:
@dangogh: is this still being worked?

dangogh:
not active on it right now,  but adding to it with other work..   It'll take some time to get all cmds covered..

smalenfant:
We are now building with docker.

PR #738, opened 2015-11-02
build all rpms -- round 2

dangogh:
I believe it's complete now based on the feedback provided in the previous PR #721.
  - All rpms build successfully with no environment vars required other than normal dev environment.
  - All subprojects get version from top-level VERSION file.
  - Each creates both a install rpm and src rpm (except traffic_monitor and traffic_router which build thru mvn).
  - traffic_ops_ort and traffic_ops are both built using the same script.
  - New subprojects are automatically included.
  - All subprojects built by default,  or specific ones provided as arguments to build/build.sh


dangogh:
This fixes #507 -- exit status is now used consistently throughout all the build scripts.

PSUdaemon:
This looks good.

If I attempt to build this, will it write anything outside of the build dirs?

dangogh:
it writes only in rpmbuild and dist directories at the top level of traffic_control

dangogh:
@PSUdaemon -- I've added a README.md as suggested..

dangogh:
you're right -- carton is not needed until postinstall.   I'll remove that.

PSUdaemon:
This looks good now, except there is a merge conflict. Can you rebase against master and I will merge?

dangogh:
@PSUdaemon rebased with master...

PR #739, opened 2015-11-03
Decrease color contrast in traffic ops web interface

trevorackerman:


knutsel:
I think we should make this an install option somehow? Not sure if we should merge this as-is.

trevorackerman:
That'd be great but would take a lot more work. We probably shouldn't merge this in if we really need to have it configurable. I'm ok with punting on it.

PR #751, opened 2015-11-05
Updated CSS for current traffic values

dneuman64:


mitchell852:
fyi - id's should be unique per page. if you have multiple things that you want to style the same, use a css class

<div class="stat_label">

dneuman64:
fair point.

PR #781, opened 2015-11-13
check for undefs when building profile cache list

### LINKED ISSUE #778: Undefined value causes exception when creating CRConfig snapshot
```Traffic Ops fatal error occurred while processing your request. 
Error at line 96 (    -in => [ @{ $profile_cache->{'CCR'} }, @{ $profile_cache->{'EDGE'} }, @{ $profile_cache->{'MID'} } ]) 
Can't use an undefined value as an ARRAY reference at /opt/traffic_ops/app/lib/UI/Topology.pm line 96.```

..in this case, `{$profile_cache->{'MID'}` did not exist.

Line 96 has this:

`-in => [ @{ $profile_cache->{'CCR'} }, @{ $profile_cache->{'EDGE'} }, @{ $profile_cache->{'MID'} } ]}}`

..any of those keys could be missing, so change the code to build the array for the `-in` argument in a more guarded manner.

dangogh:
Fixes #778

### DIFF
--- a/traffic_ops/app/lib/UI/Topology.pm
+++ b/traffic_ops/app/lib/UI/Topology.pm
@@ -90,10 +90,15 @@ sub gen_crconfig_json {
 	}
 
 	my %param_cache;
+	my @profile_caches;
+	for my $cachetype (qw/CCR EDGE MID/) {
+		my $r = $profile_cache->{$cachetype};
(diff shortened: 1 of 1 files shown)


dangogh:
Updated to throw exception if any profile cache group is empty

PR #792, opened 2015-11-17
api documentation cleanup plus adding required role to each api endpo…

mitchell852:
…int/route


dewrich:
Let's talk about the Route definition format tomorrow

mitchell852:
resource path added back per dewayne

PR #807, opened 2015-11-20
fix exec permissions in rpm; add check for rpmbuild

### LINKED ISSUE #805: TO: some .pl files are missing exec permission
Most likely rpm spec file issue.  In particular,  app/db/admin.pl is missing that permission.

dangogh:
Fixes #805


PSUdaemon:
Can you do a PR for back port to 1.3.x as well? Thanks.

PR #808, opened 2015-11-20
Fixes #806

### LINKED ISSUE #806: TO: /api/{version}/asns.json has superfluous top-level key
```{
response: {
asns: [ ... ]
}
}```

should just be:

```{
response: [ ... ]
} ```

jheitz200:


mtorluemke:
Even though this could be considered a 'bug' in the 1.1 API, I still think the new structure should only exist in version 1.2+ -- unknown how many folks are relying on the old structure.

mitchell852:
make sure you update the docs

mitchell852:
so really you don't need this anymore in the docs:

>lastUpdated

the > is used to show that the property is nested...and it no longer is now.

jheitz200:
thanks for pointing that out

PR #824, opened 2015-11-30
update download_web_deps

dangogh:
cleaner and more robust -- updates files that are different from downloaded version, creates if needed, but produces warning for any file(s) needing updating but unable to update.   Produces error(s) for any files that don't already exist and can't be written.   Works thru all files regardless and exits with an error at the end if any errors found.


smalenfant:
Does it download the files if it already has it?

dangogh:
yes -- it downloads the file and does md5sum on downloaded copy and existing copy.   If they match, it does not replace and reports "Kept ....".  If not, it replaces and reports "Replaced ...".   If it does not exist, it reports "Created...".

Issue #820, opened 2015-11-25
Content revalidate fails on updating Server update bits


smalenfant:
Once you get content revalidation SQL Table right (needs documented), there is a failure when submitting related to `Set up_pending` which seems to be log insert problem.

Who's tm_user = 0 ?

```
Traffic Ops fatal error occurred while processing your request. 
Error at line 215 (    my $insert = $self->db->resultset('Log')->create() 
DBIx::Class::Storage::DBI::_dbh_execute(): DBI Exception: DBD::mysql::st execute failed: Cannot add or update a child row: a foreign key constraint fails (`traffic_ops_db`.`log`, CONSTRAINT `fk_log_1` FOREIGN KEY (`tm_user`) REFERENCES `tm_user` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION) [for Statement "INSERT INTO log ( level, message, tm_user) VALUES ( ?, ?, ? )" with ParamValues: 0='CODEBIG', 1='Set upd_pending = 1 for all applicable caches', 2=0] at /opt/traffic_ops/app/lib/UI/Utils.pm line 215
```

dewrich:
Merged PR #820

PR #833, opened 2015-12-04
Removing references to CODEBIG

smalenfant:
Please review. I made a few assumptions about the LEVEL for current log entries (2) and removed the reference to CODEBIG in log().


knutsel:
Hey Steve - thanks, but I think this can't be right:

$user = $self->db->resultset('TmUser')->search( { username => $user } )->get_column('id')->single;

you're setting $user by searching username => $user  and user is undef at that point... 

I think you need the REMOTE_USER thing?

smalenfant:
Let me check on that... I probably need to keep the above `$user = $self->current_user()->{username};`.

smalenfant:
Fixes #820

PR #838, opened 2015-12-07
add Regional Geoblocking support (TR & Ops)

weifensh:
Support regional geoblocking based on the postal code of client IPs;
Originally designed for Canadian postal code, but could be simply modified to support other countries;
The geoblocking policy is defined by a JSON file, and TR retrieves this file from a URL configured on Ops.


limited:
Docs will be appearing here as we publish them
https://github.com/Comcast/traffic_control/wiki/Regional_Geoblocking

PR #846, opened 2015-12-07
Adding values in job_status and job_agent table

### LINKED ISSUE #845: Missing row entries for content invalidation
There is missing entries in both job_agent and job_status to get invalidation to work.

smalenfant:
Fixed #845


smalenfant:
This my first time taking on one of those migration tasks... please review carefully.

dneuman64:
This is probably better off in traffic_ops/app/db/seeds.sql.  Also, please use 'insert ignore into' instead of 'insert into'.

smalenfant:
The dummy was to create an entry in the database so the Job can be created (mandatory). Although just noticed that `id` is forced to 1 in the code... What shall we do here?

```
Traffic Ops fatal error occurred while processing your request. 
Error at line 135 (	my $insert = $self->db->resultset('Job')->create() 
DBIx::Class::Storage::DBI::_dbh_execute(): DBI Exception: DBD::mysql::st execute failed: Cannot add or update a child row: a foreign key constraint fails (`traffic_ops`.`job`, CONSTRAINT `fk_job_agent_id1` FOREIGN KEY (`agent`) REFERENCES `job_agent` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION) [for Statement "INSERT INTO job ( agent, asset_type, asset_url, entered_time, job_user, keyword, object_name, object_type, parameters, start_time, status) VALUES ( ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ? )" with ParamValues: 0=1, 1="file", 2="http://op-cox-ip-ads-prod-g.comcast.net/foo/.*", 3='2015-12-07 21:26:54', 4='60', 5="PURGE", 6=undef, 7=undef, 8="TTL:48h", 9='2015-12-07 20:29:32', 10=1] at /opt/traffic_ops/app/lib/UI/Job.pm line 135
```

PR #852, opened 2015-12-08
fixing postinstall

jheitz200:
updaing postinstall to install all of the parameters in parameter.json.
updating production db config to reflect correct prod db info


knutsel:
Shouldn't there be an 

``s/__KICKSTART-FILE__/$kickstartFilesFqn/``

somewhere?

jheitz200:
@knutsel  the substitution is happening, if you look at line number 207-220 you will see what we did

knutsel:
Oh, yea see it... Not even a need for the __placeholder__

PR #853, opened 2015-12-09
add support for regional geo-blocking

weifensh:
Support regional geo-blocking based on the postal code of client IP;
Originally designed for postal code in Canada, but could be easily modified to support other countries;
The geo-blocking policy is defined by a JSON file, and TR retrieves this file from a URL configured on Ops.

docs: https://github.com/Comcast/traffic_control/wiki/Regional_Geoblocking


limited:
I added the design doc to the Wiki page, along with some high level slides

elsloo:
Apologies for the delay on reviewing/merging this PR. I'm hoping to get this into our lab for integration testing sometime this week, so hopefully we'll have this merged in the very near future, or next week at the absolute latest.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/config/ConfigHandler.java
```diff
@@ -344,6 +353,23 @@ private void parseCoverageZoneNetworkConfig(final JSONObject config) throws JSON
...
+	 * Parses the RegionalGeo database configuration and updates the database if the URL has
+	 * changed.
+	 *
+	 * @param config
+	 *           the {@link TrafficRouterConfiguration}
+	 */
+	private void parseRegionalGeoConfig(final JSONObject config) {
+		try {
+			getRegionalGeoUpdater().setDataBaseURL(
+				config.getString("regional_geoblock.polling.url"),
```
elsloo: Can this be modified to use optString() or a conditional block that uses has() to see if regional_geoblock.polling.url exists before proceeding? The first thing I saw when I started this up in the lab was an exception generated from this section because I lack the regional_geoblock.polling.url parameter, and most users will be in the same boat unless they're using the regional geoblocking feature.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/http/TRServlet.java
```diff
@@ -161,6 +167,7 @@ private void writeHttpResponse(final HttpServletResponse response, final HttpSer
 					response.sendRedirect(location.toString());
 				}
 			}
+			httpAccessRecordBuilder.rerr(track.getResultInfo());
```
elsloo: I would prefer not to conflate the rerr field in the logs to contain what would be valid regional geo blocking results. Could you please introduce a new field to store additional data such as radd, such that it's generic and can be used for additional String data by other implementations?

I believe that we'll need to add some code to the HTTPAccessRecordBuilder, HTTPAccessRecord, HTTPAccessEvent and HTTPRouteResult classes in order to introduce the "radd" log field and a property to store the data that's accessible from TrafficRouter.route().

weifensh: Thanks Jeff for the comments. I will revise accordingly and update the pull request in one or two days.

dewrich:
@eric_friedrich: Haven’t dug too deep, but I noticed that your API keys are “snake case” ie: ‘delivery_service_id', our convention is camelCase (language neutral), so ‘deliveryServiceId’ instead, that change seems fairly pervasive if you want to correct that throughout your PR.  I still have it on my plate to review

weifensh:
Thanks Dewayne. Sure, we will follow the camelCase convention, and update the pull request soon.

dewrich:
So I just looked at the TO side of this and I didn't see any functionality to expose this through the API (along with test cases under t/api/1.x - not sure which version this will be in).  To what extent has this feature been integration tested?

dangogh:
This also now has conflicts with some changes I made recently to use of ->param(...) in UI/DeliveryService.pm.   That will need to be resolved as well.

yorma:
For TO side, there is not API added. There is modification in t/deliveryservices.t, not change test cases in t/api/1.x.  Will check conflicts, thanks.

weifensh:
Happy new year, Jeff, Dewayne and Dan. Thanks for your comments. We have revised the request based on your comments. More comments are welcome.

elsloo:
Hi; unfortunately this cannot be merged cleanly again. Could you please rebase? We're going to start reviewing the Java side again shortly.

weifensh:
Thanks Trevor for the comments on TR. I have revised the code based on your comments. So to reduce the code changes on the class TrafficRouter, the Regional Geoblocking logic has been adjusted to be executed after the cache selection logic in router(). Could you help to review the new code? Thanks.

dewrich:
@weifensh We are really close to pulling in this PR after a bit more testing, but the first problem I hit was the nature of migrations in that we've put several migrations now ahead of yours, so the way goose works it looks at the latest date/timestamp of the migration to determine whether or not to apply any changes.

So if you can rename your db/migrations/20151111100000_add_regional_geo_blocking.sql to some date AFTER the last migration which currently is "20160102193037_drop_local_user.sql", that will help the PR to come in more smoothely.

weifensh:
@dewrich Just renamed the sql file. Thanks.

PR #868, opened 2015-12-14
only admin user can view and update the ort cron job config

### LINKED ISSUE #867: Admin password visible to Operator User in TrafficOps UI
Currently if an operator user is in the Traffic Ops UI, the admin password is visible in the field that holds the cronjob information for the ort script. This poses a security risk.

zhilhuan:
fix #867

### DIFF
--- a/traffic_ops/app/lib/API/Parameter.pm
+++ b/traffic_ops/app/lib/API/Parameter.pm
@@ -31,11 +31,16 @@ sub index {
 	my $rs_data = $self->db->resultset("ProfileParameter")->search( undef, { prefetch => [ 'parameter', 'profile' ] } );
 	my @data    = ();
 	while ( my $row = $rs_data->next ) {
+        my $value = $row->parameter->value;
+        if(!&is_admin($self)){ # Prevent other users to peek admin password in cron configuration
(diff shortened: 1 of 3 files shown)


dewrich:
So, a better approach might be to add a "secure" boolean column (only modifiable to 'admins') to the parameter table and TO UI to prevent visibility to the "value".  Adding this "secure" parameter feature will allow us to utilize it for other "sensitive" parameters throughout TrafficOps

zhilhuan:
@dewrich Yes, I totally agree with you. This is a quick fix. The general solution may be adding the "secure" check box when adding/editing a parameter. So you are planing to add the "secure parameter" as a new feature, right?

dewrich:
We aren't planning on it currently, but "patches welcome". :)

dneuman64:
@dewrich @zhilhuan what is the status of this PR? It is 3 months old.  Is this something that needs to be closed or is this something that we still want to consider merging?  If the latter than the conflicts need to be resolved.

zhilhuan:
@dneuman64 We are implementing another "secure parameter" solution as dewrich suggested. Another PR will be created to replace this one. I think we can close this PR. Thanks!

Issue #863, opened 2015-12-11
The API @ /api/1.2/hwinfo.json will fail with an empty reply from the server when too many rows are returned


lwickline:
When there are too many entires in the hardware table, calls to the API will fail with an empty reply from server.

PR #869, opened 2015-12-14
Issue/863

dewrich:
Fixes: #863


dneuman64:
In the future we should make better pull request titles so we can know at a glance what is changing.  Otherwise we have to go to the PR itself and look.

PR #880, opened 2015-12-15
adds secondary_parent=... to parent.config

dangogh:


dangogh:
does not yet provide ability to add secondary parent cachegroups thru UI

Issue #859, opened 2015-12-10
Traffic Ops hardware Tab fails to populate when there are too many rows returned


lwickline:
The API @ /api/1.2/hwinfo.json will fail with an empty reply from the server. 

The hardware page just never loads. 

example curl:

time curl -H 'Cookie: mojolicious=blah' -Lvsk "https://server.server.comt/api/1.2/hwinfo.json" -o /dev/null
*   Trying 1.2.3.4...
* Connected to server.server.com (1.2.3.4) port 443 (#0)
* TLS 1.2 connection using TLS_RSA_WITH_AES_128_GCM_SHA256
* Server certificate: server.server.com
* Server certificate: COMODO RSA Organization Validation Secure Server CA
* Server certificate: COMODO RSA Certification Authority
* Server certificate: AddTrust External CA Root
> GET /api/1.2/hwinfo.json HTTP/1.1
> Host: tm.comcast.net
> User-Agent: curl/7.43.0
> Accept: */*
> Cookie: mojolicious=blah
> 
* Empty reply from server

dewrich:
Fixed by PR: #859

PR #890, opened 2015-12-23
Solves #859 Traffic Ops hardware Tab fails to populate when there are too many rows returned

dewrich:
https://github.com/Comcast/traffic_control/issues/859

Added pagination so that all of the rows aren't pulled back at once, should probably be retrofitted to other heavy pages


traffic_ops/app/lib/API/HwInfo.pm
```diff
@@ -44,4 +44,91 @@ sub index {
...
+	my $sort_order     = $self->param("sSortDir_0") || "asc";
+	my $search_field   = $self->param("sSearch");
+	my $sort_column    = $self->param("iSortCol_0");
+
+	# NOTE: If changes are made to send additional columns then this mapping has to be updated
+	# to match the Column Number coming from datatables to it's name
+	# Unfortunately, this is a short coming with the jquery datatables ui widget in that it expects
+	# an array arrays instead of an array of hashes
+	$self->app->log->debug( "sort_order #-> " . $sort_order );
+	my $sort_direction = sprintf( "-%s", $sort_order );
```
dangogh: this can be simply "-$sort_order" -- it's much more concise.

traffic_ops/app/lib/API/HwInfo.pm
```diff
@@ -44,4 +44,91 @@ sub index {
...
+	$self->app->log->debug( "sort_direction #-> " . $sort_direction );
+	my %column_number_to_name = ( 1 => 'serverid.host_name', 2 => 'description', 3 => 'val', 4 => 'last_updated' );
+	my $column_name = $column_number_to_name{$sort_column} || "serverid";
```
dangogh: array is probably better here:
my @column_number_to_name = qw{ serverid.host_name description val last_updated };
my $column_name = $column_number_to_name[$sort_column-1] || "serverid";

Then a newly inserted column would just shift everything else to the right

traffic_ops/app/lib/UI/Cdn.pm
```diff
@@ -437,26 +438,85 @@ sub adeliveryservice {
...
+	my $sort_column    = $self->param("iSortCol_0") || "id";
+	my $echo           = $self->param("sEcho");
+
+	# NOTE: If changes are made to send additional columns then this mapping has to be updated
+	# to match the Column Number coming from datatables to it's name
+	# Unfortunately, this is a short coming with the jquery datatables ui widget in that it expects
+	# an array arrays instead of an array of hashes
+	my $sort_direction = sprintf( "-%s", $sort_order );
+	my %column_number_to_name = ( 1 => 'serverid.host_name', 2 => 'description', 3 => 'val', 4 => 'last_updated' );
+	my $column_name = $column_number_to_name{$sort_column} || "serverid";
```
dangogh: same comments as above

PR #892, opened 2015-12-28
Creates delivery service request api

mitchell852:
which captures a handful of values and emails them to the specified email address to initiate the delivery service provisioning process


mitchell852:
will open this after the break

PR #897, opened 2016-01-04
check profile name when create/update

### LINKED ISSUE #896: Health params.json updated with certain Profile names
Create Profile names in the Traffic_OPs with names such as Edge and Mid for Edge and Mid cache profiles, these don't get updated in the heath_params.json on the Traffic monitor.

In the current implementation of traffic ops, there is no "type" for a profile. When generating the health-params.json for traffic monitor, ops do not know which profile is for edge or middle or router, etc. So in traffic ops code, it use the convention that:
profile starting with "RASCAL" is a traffic monitor profile
profile starting with "CCR" is a traffic router profile
profile starting with "EDGE" is an edge server profile
profile starting with "MID" is a middle server profile

It is better to add a type for each profile, or make a restriction for the profile name.

zhilhuan:
fix #896

### DIFF
--- a/traffic_ops/app/lib/UI/Profile.pm
+++ b/traffic_ops/app/lib/UI/Profile.pm
@@ -141,7 +141,7 @@ sub check_profile_input {
 	$self->field('profile.name')->is_required;
 	$self->field('profile.description')->is_required;
 
-	if( !($name =~ m/^CCR|^EDGE|^GLOBAL|^MID|^ORG|^RASCAL|^RIAK|^TRSTATS/) ) {
+	if( !($name =~ m/^TR|^EDGE|^GLOBAL|^MID|^ORG|^TM|^RIAK|^TRSTATS/) ) {
 		$self->field('profile.name')->is_equal( "", "profile name must start with CCR|EDGE|GLOBAL|MID|ORG|RASCAL|RIAK|TRSTATS");
 	}
 
--- a/traffic_ops/app/lib/UI/Profile.pm
+++ b/traffic_ops/app/lib/UI/Profile.pm
@@ -140,6 +140,11 @@ sub check_profile_input {
 	#Check required fields
 	$self->field('profile.name')->is_required;
 	$self->field('profile.description')->is_required;
+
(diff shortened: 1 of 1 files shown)


traffic_ops/app/lib/UI/Profile.pm
```diff
@@ -140,6 +140,11 @@ sub check_profile_input {
 	#Check required fields
 	$self->field('profile.name')->is_required;
 	$self->field('profile.description')->is_required;
+
+	if( !($name =~ m/^CCR|^EDGE|^GLOBAL|^MID|^ORG|^RASCAL|^RIAK|^TRSTATS/) ) {
```
dewrich: What happens when we add another type to the "types" table, are these "hard coded" types the only ones we ever care about?

zhilhuan: I think we need add the new type here. 

The original problem is if a profile created with name like "Edge-1" or "edge1", it will not be considered as an edge profile in traffic_ops/app/lib/MojoPlugins/Health.pm. And the user will not be alerted. 

So it is better to notify users that only the predefined profile types are allowed, either by add a "type" property to the "Profile" table, or make a validation for profile name like this fix.

smalenfant:
@zhilhuan From the Documentation, RASCAL is now TM and CCR now TR. 

http://traffic-control-cdn.net/docs/latest/admin/traffic_ops_config.html

dneuman64:
While I like the idea, we have several profiles that would not work with the regex  `if( !($name =~ m/^TR|^EDGE|^GLOBAL|^MID|^ORG|^TM|^RIAK|^TRSTATS/) ) {` for examples, ELASTICSEARCH and KAFKA.  I don't know that we want to have to keep this list updated somewhere.  I think we should probably table this until we can figure out a better solution (probably in 2.0).  If you agree, I will close.

zhilhuan:
@dneuman64 Yes, I agree.

dneuman64:
Closing. This is a problem that will need to be addressed in 2.0.

Issue #904, opened 2016-01-05
Missing TTL for DNSSEC Key Causes JSONException


elsloo:
If a DNSSEC key is missing a TTL, or its TTL is null, a JSONException is thrown in Traffic Router. The TTL shouldn't be missing, and we should have defensive code in TR to handle this scenario.

PR #908, opened 2016-01-05
Fixes for RuntimeExceptions in maintenance Runnables, NPE caused by missing TTL on DNSSEC keys, only create SignedZoneKeys for Names that have keys, and fixed an oversight where we weren't converting to millis when calculating next refresh for a given SignedZoneKey.

elsloo:


dneuman64:
closes #904

PR #916, opened 2016-01-07
lock on influxdb v0.9.4 api

dangogh:


dangogh:
accidentally pulled in another change -- don't pull this in yet...

PR #917, opened 2016-01-07
lock on influxdb v0.9.4 api

### LINKED ISSUE #875: Build process shouldn't 'go get' the influxdb client
It looks like when we are building traffic stats, we do a `go get -v`  which appears to have caused issues with influxdb. With traffic stats 1.3.0 RC3, when data is written to infuxdb (0.9.4.2) it has an incorrect timestamp.  I have seen this before and it was related to using a version of the client that was ahead of the currently installed influxdb version.    I know there isn't a good way to tell `go get` to use a certain version, but maybe we should handle this differently?  The Influxdb client appears to be backward compatible (the 0.9.3 version of the client works with 0.9.4.2) so maybe we should handle this manually and identify a minimum version supported.

dangogh:
Fixes #875

### DIFF
--- a/traffic_stats/build/traffic_stats.spec
+++ b/traffic_stats/build/traffic_stats.spec
@@ -25,7 +25,7 @@ Packager:	david_neuman2 at Cable dot Comcast dot com
 Vendor:		Comcast Cable
 Group:		Applications/Communications
 License:	Apache License, Version 2.0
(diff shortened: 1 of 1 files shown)


dneuman64:
Looks good.  Can you cherry-pick this to the 1.3.x branch for RC4?

PR #941, opened 2016-01-19
Fixes #928 - Checking for ldap users

### LINKED ISSUE #928: /api/$version/user/current.json is not compatible with ldap
If a user logs in using ldap credentials, the api route (/api/$version/user/current.json) fails as it looks for the user in the tm_user table.

When you login via ldap a fake user is apparently created. This fake user should be returned from /api/version/user/current.json. I suppose this fake user would only include user.username as that is the only value that is known.

In addition, POST /api/$version/user/current/update should return should include some error alert along the lines of "cannot update ldap user"

jheitz200:


traffic_ops/app/lib/API/User.pm
```diff
@@ -146,31 +146,59 @@ sub current {
...
-				"country"         => $row->country,
+				"id"              => "",
+				"username"        => $current_username,
+				"role"            => $role,
+				"uid"             => "",
+				"gid"             => "",
+				"company"         => "",
+				"email"           => "",
+				"fullName"        => "",
+				"newUser"         => "",
```
mitchell852: this should be false

traffic_ops/app/lib/API/User.pm
```diff
@@ -146,31 +146,59 @@ sub current {
...
+				"localUser"       => "",
```
mitchell852: localUser should be false

jheitz200: I am going to remove localUser from the database and add a local variable called localUser

mitchell852:
i'll merge this PR once i pull jackie's branch and give it a quick test

Issue #946, opened 2016-01-20
TO: Not listening on IPv6


smalenfant:
I believe this was fixed previously.

For Traffic Ops to listen on both IPv4 and IPv6, The `https://*:443` needs changed to `https://[::]:443` 

Somehow the cdn.conf.backup1 file is correct... Which leads me to believe that something is altering the line on disk...

New generated file

```
  hypnotoad => {
    group => 'trafops',
    pid_file => '/var/run/traffic_ops.pid',
    listen => [
      'https://*:443?cert=/etc/pki/tls/certs/localhost.crt&key=/etc/pki/tls/private/localhost.key&ca=/etc/pki/tls/certs/localhost.ca&verify=0x00&ciphers=AES128-GCM-SHA256:RC4:HIGH:!MD5:!aNULL:!EDH:!ED'
    ],
    workers => 24,
    user => 'trafops'
  },
```

Backup file 

```
[root@cms01 install]# cat /opt/traffic_ops/app/conf/cdn.conf.backup1
{
    hypnotoad => {
        listen => [
            'https://[::]:443?cert=/etc/pki/tls/certs/localhost.crt&key=/etc/pki/tls/private/localhost.key&verify=0x00&ciphers=AES128-GCM-SHA256:RC4:HIGH:!MD5:!aNULL:!EDH:!ED'
        ],
        user     => 'trafops',
        group    => 'trafops',
        pid_file => '/var/run/traffic_ops.pid',
        workers  => 48
    },
```

File on disk

```
-rw-r--r--. 1 trafops trafops  836 Jan 20 13:35 cdn.conf
-rw-r--r--. 1 trafops trafops 1344 Jan 20 01:05 cdn.conf.backup1
-rw-r--r--. 1 trafops trafops  804 Jan 20 04:41 cdn.conf.backup2
```

smalenfant:
Found where this is getting overwritten.. 

```
sub writeCdn_conf {
        my $cdn_conf = shift;

        # listen param to be inserted
        my $listen_str = "https://*:443?cert=${cert}&key=${key}&ca=${ca}&verify=0x00&ciphers=AES128-GCM-SHA256:RC4:HIGH:!MD5:!aNULL:!EDH:!ED";
```

limited:
Jeff and Steve say to close it

PR #947, opened 2016-01-20
Enable Traffic Ops for IPv6

smalenfant:
genererateCert was overwriting the listen line with non-IPv6 values.


smalenfant:
Fixes #946

knutsel:
doesn't this make it listen _only_ on IPv6?

Issue #938, opened 2016-01-18
TO: missing good.csv on updateservercsv


dangogh:
The Wiki tab on the uploadservercsv page mentions a sample "good.csv",  but that file is not available.

dewrich:
See me on this one, I found the good.csv in a different directory

dewrich:
I fixed this, see PR #952

dewrich:
Added back good.csv through PR #952

PR #952, opened 2016-01-21
Fix for Issue #938 - added back good.csv

dewrich:


dangogh:
it's missing the one column:  CACHEGROUP entry between LOCATION and PHYS_LOCATION.

PR #966, opened 2016-01-22
Updated Traffic Stats to use v2 version of Influxdb Client

### LINKED ISSUE #956: Upgrade Traffic Stats to use v2 client for InfluxDb
The v1 client for influxdb will soon be deprecated, so we need to upgrade all the programs in Traffic Stats to use the v2 client.  See [here](https://github.com/influxdata/influxdb/tree/master/client) for more information.
>**NOTE**: The Go client library now has a "v2" version, with the old version being deprecated. The new version can be imported at import "github.com/influxdb/influxdb/client/v2". It is not backwards-compatible.

This will be a breaking change for Traffic Stats in that Traffic Stats will require **at least** version 0.9.6 of InfluxDb to support the v2 client.

dneuman64:
This closes #956.

InfluxDb will need to be upgraded to version 0.9.6 or above to support the client change.

### DIFF
--- a/traffic_stats/traffic_stats.go
+++ b/traffic_stats/traffic_stats.go
@@ -27,7 +27,6 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"net/http"
-	"net/url"
 	"os"
 	"os/signal"
 	"strconv"
@@ -700,13 +699,10 @@ func getURL(url string) ([]byte, error) {
 
 func influxConnect(config StartupConfig, runningConfig RunningConfig) (influx.Client, error) {
 	// Connect to InfluxDb
-	var urls []*url.URL
+	var urls []string
 
 	for _, InfluxHost := range runningConfig.InfluxDBProps {
-		u, err := url.Parse(fmt.Sprintf("http://%s:%d", InfluxHost.Fqdn, InfluxHost.Port))
-		if err != nil {
-			continue
-		}
+		u := fmt.Sprintf("http://%s:%d", InfluxHost.Fqdn, InfluxHost.Port)
 		urls = append(urls, u)
 	}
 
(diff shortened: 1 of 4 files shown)


traffic_stats/traffic_stats.go
```diff
@@ -713,19 +715,13 @@ func influxConnect(config StartupConfig, runningConfig RunningConfig) (*influx.C
 		url := urls[n]
 		urls = append(urls[:n], urls[n+1:]...)
 
-		conf := influx.Config{
-			URL:      *url,
+		conf := influx.HTTPConfig{
+			Addr:     url.String(),
```
PSUdaemon: Since the client no longer takes a URL struct, you can probably skip the Parse() and String() steps. They are completely superfluous now.

traffic_stats/traffic_stats.go
```diff
@@ -748,29 +744,12 @@ func sendMetrics(config StartupConfig, runningConfig RunningConfig, bps influx.B
...
-		_, err = influxClient.Write(chunk_bps)
-		if err != nil {
-			if retry {
-				config.BpsChan <- chunk_bps
-			}
-			errHndlr(err, ERROR)
-		} else {
-			log.Debug("Sent ", len(chunk_bps.Points), " stats")
-		}
-	}
```
PSUdaemon: Sad to see this code go (I felt it rather clever), but glad they fixed their client.

traffic_stats/traffic_stats.go
```diff
@@ -748,29 +744,12 @@ func sendMetrics(config StartupConfig, runningConfig RunningConfig, bps influx.B
...
+	log.Info(fmt.Sprintf("Sent %v stats for %v", len(bps.Points()), bps.Database()))
 }
 
+//IntMin returns the lesser of two ints
```
PSUdaemon: golint...:trollface:

PR #971, opened 2016-01-25
New feature to retrieve all of the Delivery Services with their regexes

dewrich:


mitchell852:
Can you get the regexes for just one deliveryservice as opposed to all delivery services? if so, i'd call this route: /api/{version}/deliveryservice_regexes.json instead of /api/{version}/deliveryservices_regexes.json

PR #991, opened 2016-01-26
Add Traffic Ops postinstall default parameters

### LINKED ISSUE #855: Clean install profiles are out of date and incomplete.
The profiles for edges are ATS 520, the mid is 421 (from the name). The mid profile has only 4 entries in it. 

Let's keep this open until I get further into the 1.3 test install until we try to fix it.

### LINKED ISSUE #977: TO: write script to import/export parameters from TO PostInstall
As a developer I need to write a script to import/export all parameters from TO so that I can get postinstall working with a fresh install.

Acceptance Criteria:
- Parameters available for use in PostInstall
- Docs

### LINKED ISSUE #990: Traffic Ops postinstall fails to add user to SQL
With this issue, after installing TrafficOps and running `postinstall` you can't log in to Traffic Ops, unless you manually insert the user into `tm_user` in SQL.

The issue is an extra parameter in `dataload.go`, should be easy and safe to fix.

robert-butts:
This adds a `get_parameters.go` script which gets profiles from a live Traffic Ops instance and formats them as `.traffic_ops` files to be imported by TO, replacing CDN-specific data with template variables.

This also adds default profiles, and adds functions to `postinstall` to replace the template values with user input and upload those profiles to the new TO.

This also fixes a bug in `dataload.go` which caused `postinstall` to fail to insert the TO user in SQL.

Fixes #855, fixes #977, fixes #990.


traffic_ops/install/bin/postinstall
```diff
@@ -432,16 +435,184 @@ else {
 	$rc = execCommand( "./download_web_deps", "-i" );
 	if ( $rc != 0 ) {
 		print "ERROR: failed to install Traffic Ops Web dependencies, check the console output and rerun postinstall once you've resolved the error.\n";
-		exit 5;
 	}
 }
 
+sub profile_replace {
+		my($profile) = @_;
+		open(FH, "<$profile") or die ("open(): $!");
```
dangogh: rather than use FH,  pass a scalar variable to open.  If it's undef, the fh value will be placed in it.   Also,  always use the 3 arg form of open:
open(my $fh, '<', $profile) or die ...

traffic_ops/install/bin/postinstall
```diff
@@ -432,16 +435,184 @@ else {
...
+		my @lines = <FH>;
```
dangogh: you're slurping in the whole file here and then making a modified copy of all the lines.   Why not just go thru the lines and store only the modified version?

```while (<$fh>) {   s/{{...}}/.../g;   push @newlines, $_; }```

File: traffic_ops/install/bin/postinstall [hunk 3]
```diff
@@ -432,16 +435,184 @@ else {
...
+				$_ =~ s/{{.HealthThresholdLoadavg}}/$parameters->{"health.threshold.loadavg"}/g;
+				$_ =~ s/{{.HealthThresholdAvailableBandwidthInKbps}}/$parameters->{"health.threshold.availableBandwidthInKbps"}/g;
+				$_ =~ s/{{.RAMDrivePrefix}}/$parameters->{"RAM_Drive_Prefix"}/g;
+				$_ =~ s/{{.RAMDriveLetters}}/$parameters->{"RAM_Drive_Letters"}/g;
+				$_ =~ s/{{.HealthConnectionTimeout}}/$parameters->{"health.connection.timeout"}/g;
+				$_ =~ s\{{.CronOrtSyncds}}\*/15 * * * * root /opt/ort/traffic_ops_ort.pl syncds warn $parameters->{'tm.url'} $tmAdminUser:$tmAdminPw > /tmp/ort/syncds.log 2>&1\g;
+				push(@newlines,$_);
+		}
+
+		open(FH, ">$profile") or die ("open(): $!");
```
dangogh: what happens to the profile if something kills it immediately after this open for write?

traffic_ops/install/bin/postinstall
```diff
@@ -432,16 +435,184 @@ else {
...
+		  $parameters->{"health.threshold.loadavg"} = InstallUtils::promptUser ("Health Threshold Load Average", "25");
+	  }
+	  while (!defined $parameters->{"health.threshold.availableBandwidthInKbps"} || $parameters->{"health.threshold.availableBandwidthInKbps"} eq "" || $parameters->{"health.threshold.availableBandwidthInKbps"} eq ">") {
+		  $parameters->{"health.threshold.availableBandwidthInKbps"} = ">" . InstallUtils::promptUser ("Health Threshold Available Bandwidth in Kbps", "1750000");
+	  }
+	  while (!defined $parameters->{"health.connection.timeout"} || $parameters->{"health.connection.timeout"} eq "") {
+			$parameters->{"health.connection.timeout"} = InstallUtils::promptUser ("Traffic Server Health Connection Timeout (milliseconds)", "2000");
+	  }
+
+		profile_replace("/opt/traffic_ops/install/bin/profile.global.traffic_ops");
```
dangogh: is there a better place to store the profiles than the /bin directory?

traffic_ops/install/bin/postinstall
```diff
@@ -432,16 +435,184 @@ else {
...
+# Returns the cookie, or the empty string on error
+sub get_traffic_ops_cookie {
+		my($uri, $user, $pass) = @_;
+
+		my $loginUri = "/api/1.2/user/login";
+
+		my $curl = WWW::Curl::Easy->new;
+		my $response_body = "";
+		open(my $fileb, ">", \$response_body);
+		my $loginData = "{ \"u\":\"$user\", \"p\":\"$pass\" }";
```
dangogh: this is where the qq{} quoting operator is nice...
```
my $loginData = qq/{"u": "$user", "p": "$pass"}/;
```

traffic_ops/install/bin/postinstall
```diff
@@ -432,16 +435,184 @@ else {
...
+		$curl->setopt(WWW::Curl::Easy::CURLOPT_SSL_VERIFYPEER, 0);
+		$curl->setopt(WWW::Curl::Easy::CURLOPT_HEADER, 1); # include header in response
+		$curl->setopt(WWW::Curl::Easy::CURLOPT_NOBODY, 1); # disclude body in response
+		$curl->setopt(WWW::Curl::Easy::CURLOPT_POST, 1);
+		$curl->setopt(WWW::Curl::Easy::CURLOPT_POSTFIELDS, $loginData);
+		$curl->setopt(WWW::Curl::Easy::CURLOPT_WRITEDATA, $fileb);	# put response in this var
+		$curl->perform();
+
+		my $cookie = $response_body;
+		if($cookie =~ /mojolicious=(.*); expires/)
```
dangogh: did you run perltidy on this?

traffic_ops/install/bin/postinstall
```diff
@@ -432,16 +435,184 @@ else {
...
+		my $toUser = $tmAdminUser;
+		my $toPass = $tmAdminPw;
+
+		my $toCookie = get_traffic_ops_cookie($toUri, $toUser, $toPass);
+
+		print "Got cookie: " . $toCookie;
+
+		# \todo use an array?
+		print "Importing Global profile...\n";
+		profile_import_single("/opt/traffic_ops/install/bin/profile.global.traffic_ops", $toUri, $toCookie);
```
dangogh: wouldn't it be better to pass in to import_profiles the directory where the profile data is stored rather than hard-coding it many times here?

traffic_ops/install/go/src/comcast.com/dataload/dataload.go
```diff
@@ -408,7 +408,7 @@ func loadUsers(db *sql.DB, dbName string) error {
 			return err
 		}
```
dangogh: There are 2 separate changes in this file -- the former should be a completely separate pull request applied to issue #990.  The latter part should be part of this one...

traffic_ops/install/bin/postinstall
robert-butts: What about `install/data/json`?

dangogh: that would make more sense, but would that interfere with way the other json files there are used?  what are they used for?   what about a new install/data/ subdirectory?

robert-butts: It's probably better to treat the profiles as black boxes anyway, so everything theoretically works if Traffic Ops import is changed from JSON to something else. `install/data/profiles`?

dangogh: that works for me...

PR #997, opened 2016-01-28
Fix formatting of lat,long in traffic router access log 

trevorackerman:
and log an error if a cache group with a location of 0.0,0.0 is selected during Geo lookup

(cherry picked from commit a06b20443078c19a4726104b88af3c38f4cfba1d)


trevorackerman:
already taken care of from previous PR's

elsloo:
Because RC0 is up for a vote, this is a nice to have for 1.4.x if we are planning to do an RC1.

PR #1007, opened 2016-02-04
Put chunking back in to address OOM issues

### LINKED ISSUE #1004: Traffic Stats 1.4.0RC0 - Out of memory issues
Traffic Stats 1.4.0 is running out of memory and getting killed.  It appears that Traffic Stats gets an error message back and then stops being able to connect to Traffic Monitor.  Although the logging does not show a smoking gun (need to fix that), my belief is that this has to do with taking out the "chunking" functionality when the client was upgraded.

dneuman64:
this closes #1004


PSUdaemon:
Glad to see chunking return!

dneuman64:
:+1:

PR #1034, opened 2016-02-11
This updates traffic stats to use the new github repo for influxdb (github.com/influxdata)

dneuman64:
The spec file will now build with the master branch of InfluxDb, because I could not getting working with a specific branch due to the github repo moving.  I suspect that once a new tag is created by influxdb we can make this work again.


PSUdaemon:
Someone should tell them to push tags too...

PR #1039, opened 2016-02-11
Traffic Monitor cleanup DataSummary class, add tests for it

trevorackerman:


trevorackerman:
Wicket uses reflection so a lot of this code isn't dead after all.

PR #1052, opened 2016-02-15
Workaround for ORT plugin without params

### LINKED ISSUE #968: ORT: Raw remap text concatenated and plugin verification fails
This feature was working in Traffic Ops 0.53c. Doesn't seem to work after 0.55a.

To apply an ACL on a delivery service, I add the following in the `Raw Remap Text`

`@action=allow @src_ip=192.168.0.0-192.168.0.254`

Although during the processing, the plugin verification seems to fail.

```
Version of this script: 0.55a
WARN Sleeping for 1 seconds: .
ERROR Traffic Ops is signaling that an update is waiting to be applied.
WARN Sleeping for 12 seconds: ............
ERROR Deleting directories older than 1452886674
WARN URL: https://cdn1cdcms01.coxlab.net/genfiles/view/psp6cdatsec05/cache.config returned only the header.
ERROR Package for plugin: cache_range_requests.so@action=allow@src_ip=192.168.0.0-192.168.0.254 is not installed
ERROR Package for plugin: cache_range_requests.so@action=allow@src_ip=192.168.0.0-192.168.0.25 is not installed
ERROR Prereqs failed for remap.config!
ERROR Package for plugin: cache_range_requests.so@action=allow@src_ip=192.168.0.0-192.168.0.254 is not installed
ERROR Package for plugin: cache_range_requests.so@action=allow@src_ip=192.168.0.0-192.168.0.25 is not installed
ERROR Prereqs failed for remap.config!
ERROR Traffic Ops requires an update, but applying the update locally failed. Traffic Ops is not being updated!
```

smalenfant:
Fixes #968. The ORT script has a problem when a plugin doesn’t have
@pparam. This fixes the problem using `cache_range_request` and ACL
(@allow).

@mtorluemke Here's a good one for you.. I checked with @jrushf1239k and his plugin doesn't care about params. This is basically a workaround to ORT parsing.

### DIFF
--- a/traffic_ops/app/lib/UI/ConfigFiles.pm
+++ b/traffic_ops/app/lib/UI/ConfigFiles.pm
@@ -863,7 +863,7 @@ sub remap_dot_config {
 				$mid_remap{ $remap->{org} } .= " \@plugin=cacheurl.so \@pparam=" . $remap->{cacheurl_file};
 			}
 			if ( $remap->{range_request_handling} == 2 ) {
-				$mid_remap{ $remap->{org} } .= " \@plugin=cache_range_requests.so";
+				$mid_remap{ $remap->{org} } .= " \@plugin=cache_range_requests.so \@pparam=none";
 			}
 		}
 		foreach my $key ( keys %mid_remap ) {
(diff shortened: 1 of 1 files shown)


knutsel:
@mtorluemke, @smalenfant - wouldn't it be better to fix the ort script so it works with single plugin and no pparam? (we have the cache_range_request in our config without pparam, but have other remap plugins on the same line, that seems to work...)

smalenfant:
@knutsel It works until you put some `Raw remap text`.  Look at #968.

knutsel:
still, better place to fix is ort script? or am i mistaken?

mtorluemke:
I agree -- ORT is broken and should be fixed.

mtorluemke:
I'll work on fixing ORT for this.

PR #1054, opened 2016-02-15
Add cache type to Traffic Stats and Traffic Monitor; fixes a config file path issue and reduces logging.

elsloo:
…es #1053.

### DIFF
--- a/traffic_stats/traffic_stats.go
+++ b/traffic_stats/traffic_stats.go
@@ -77,8 +77,8 @@ type StartupConfig struct {
 // RunningConfig is used to store runtime configuration for Traffic Stats.  This includes information
 // about caches, cachegroups, health urls, and online InfluxDB servers
 type RunningConfig struct {
-	HealthUrls      map[string]map[string]string // the 1st map key is CDN_name, the second is DsStats or CacheStats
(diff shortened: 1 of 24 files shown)


traffic_stats/traffic_stats.go
```diff
@@ -77,8 +77,8 @@ type StartupConfig struct {
 // RunningConfig is used to store runtime configuration for Traffic Stats.  This includes information
 // about caches, cachegroups, health urls, and online InfluxDB servers
 type RunningConfig struct {
-	HealthUrls      map[string]map[string]string // the 1st map key is CDN_name, the second is DsStats or CacheStats
-	CacheGroupMap   map[string]string            // map hostName to cacheGroup
+	HealthUrls      map[string]map[string]string  // the 1st map key is CDN_name, the second is DsStats or CacheStats
+	CacheMap        map[string]traffic_ops.Server // map hostName to cache
```
PSUdaemon: Consider making this a type?

elsloo: Why would making this a type help?

PSUdaemon: It would be cleaner.

PR #1079, opened 2016-02-19
Change Traffic Ops volume.config to sum to 100%

### LINKED ISSUE #1078: Change Traffic Ops volume.config so percentages to sum to 100%, and add SSD volume support
Currently, volume.config is hard-coded to set each volume to 1%. This seems to work, but the ATS documentation and source suggest they ought to be equally distributed and sum to 100%. Further, setting volumes at 1% requires the volumes be very large, which makes testing and development on small machines difficult.

While we're in the area, we also want to add support for SSD volumes, via `SSD_Volume`,`SSD_Drive_Prefix`, and `SSD_Drive_Letters` parameters, and make the code extensible so adding other volume types in the future is easy.

This is somewhat dangerous and should be thoroughly tested before being put in production. The ATS documentation is somewhat ambiguous, and appearances can be deceiving, especially in C.

robert-butts:
This is moderately dangerous and should be thoroughly tested.
ATS documentation and code suggests volumes should sum to 100%,
and nothing suggests the current 1% setting is the designed
usage. But, changing how ATS perceives volumes is a bit uncertain.

This changes volumes to equally sum to 100%. E.g., two volumes are set
to 50% each. Three volumes are 33% each.

Also:

 - Make Traffic Ops volume.config generation add SSD volume, if
`SSD_Volume`, `SSD_Drive_Prefix`, and `SSD_Drive_Letters` exist.

- Make Traffic Ops storage.config add SSD volume path and number.

Resolves #1078

### DIFF
--- a/traffic_ops/app/lib/UI/ConfigFiles.pm
+++ b/traffic_ops/app/lib/UI/ConfigFiles.pm
@@ -691,26 +691,54 @@ sub generic_config {
 	return $text;
 }
 
-sub volume_dot_config {
-	my $self = shift;
-	my $id   = shift;
-	my $file = shift;
+sub get_num_volumes {
+    my $data = shift;
 
-	my $server = $self->server_data($id);
-	my $data   = $self->param_data( $server, "storage.config" );
-	my $text   = $self->header_comment( $server->host_name );
-	if ( defined( $data->{RAM_Drive_Prefix} ) ) {
-
-		# TODO JvD: More vols??
-		$text .= "# 12M NOTE: This is running with forced volumes - the size is irrelevant\n";
-		$text .= "volume=" . $data->{RAM_Volume} . " scheme=http size=1%\n";
-		$text .= "volume=" . $data->{Disk_Volume} . " scheme=http size=1%\n";
-	}
-	else {
-		$text .= "volume=1 scheme=http size=100%\n";
-	}
-	return $text;
+    my $num            = 0;
+    my @drive_prefixes = qw( Drive_Prefix SSD_Drive_Prefix RAM_Drive_Prefix);
+    foreach my $pre (@drive_prefixes) {
+        if ( exists $data->{$pre} ) {
+            $num++;
+        }
+    }
+    return $num;
 }
+
+sub volume_dot_config_volume_text {
+    my $volume      = shift;
+    my $num_volumes = shift;
+    my $size        = int( 100 / $num_volumes );
+    return "volume=$volume scheme=http size=$size%\n";
+}
+
+sub volume_dot_config {
+    my $self = shift;
+    my $id   = shift;
+    my $file = shift;
+
+    my $server = $self->server_data($id);
+    my $data   = $self->param_data( $server, "storage.config" );
+    my $text   = $self->header_comment( $server->host_name );
+
+    my $num_volumes = get_num_volumes($data);
+
+    $text .=
+"# 12M NOTE: This is running with forced volumes - the size is irrelevant\n";
+    if ( defined( $data->{Drive_Prefix} ) ) {
+        $text .=
+          volume_dot_config_volume_text( $data->{Disk_Volume}, $num_volumes );
+    }
+    if ( defined( $data->{RAM_Drive_Prefix} ) ) {
+        $text .=
+          volume_dot_config_volume_text( $data->{RAM_Volume}, $num_volumes );
+    }
+    if ( defined( $data->{SSD_Drive_Prefix} ) ) {
+        $text .=
+          volume_dot_config_volume_text( $data->{SSD_Volume}, $num_volumes );
+    }
+    return $text;
+}
+
 
 sub hosting_dot_config {
 	my $self = shift;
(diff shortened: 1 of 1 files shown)


traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -691,25 +691,45 @@ sub generic_config {
...
+
+    my $server = $self->server_data($id);
+    my $data   = $self->param_data( $server, "storage.config" );
+    my $text   = $self->header_comment( $server->host_name );
+
+    # TODO JvD: More vols??
+    my $num_volumes =
+      ( defined( $data->{Drive_Prefix} )     ? 1 : 0 ) +
+      ( defined( $data->{RAM_Drive_Prefix} ) ? 1 : 0 ) +
+      ( defined( $data->{SSD_Drive_Prefix} ) ? 1 : 0 );
```
robert-butts: I wasn't sure if there's a better way to do this

traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -691,25 +691,45 @@ sub generic_config {
...
+    my $self = shift;
+    my $id   = shift;
+    my $file = shift;
+
+    my $server = $self->server_data($id);
+    my $data   = $self->param_data( $server, "storage.config" );
+    my $text   = $self->header_comment( $server->host_name );
+
+    # TODO JvD: More vols??
+    my $num_volumes =
```
robert-butts: I wasn't sure if there's a better way to do this

traffic_ops/app/lib/UI/ConfigFiles.pm
robert-butts: Wasn't sure if there's a better way to do this

dangogh: perhaps something like this? ```my $numvol = scalar(grep { /Drive_Prefix/ } keys %$data);```
or 
> my $numvol = 0;
> foreach my $k (keys %$data) {
>    if ($k =~ /Drive_Prefix/) {
>       $numvol++;
>    }
> }

traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -691,25 +691,45 @@ sub generic_config {
...
+      ( defined( $data->{RAM_Drive_Prefix} ) ? 1 : 0 ) +
+      ( defined( $data->{SSD_Drive_Prefix} ) ? 1 : 0 );
+    $text .=
+"# 12M NOTE: This is running with forced volumes - the size is irrelevant\n";
+    if ( defined( $data->{Drive_Prefix} ) ) {
+        text .= volume_dot_config_volume_textlume_text( $data->{Disk_Volume},
+            $num_volumes );
+    }
+    if ( defined( $data->{RAM_Drive_Prefix} ) ) {
+        num $text .=
```
dangogh: I think the ```num $text...``` should give you a syntax error..

traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -691,25 +691,45 @@ sub generic_config {
...
+          volume_dot_config_volume_text( $data->{RAM_Volume}, $num_volumes );
+    }
+    if ( defined( $data->{SSD_Drive_Prefix} ) ) {
+        num $text .=
```
dangogh: same here

traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -691,25 +691,45 @@ sub generic_config {
...
+    my $text   = $self->header_comment( $server->host_name );
+
+    # TODO JvD: More vols??
+    my $num_volumes =
+      ( defined( $data->{Drive_Prefix} )     ? 1 : 0 ) +
+      ( defined( $data->{RAM_Drive_Prefix} ) ? 1 : 0 ) +
+      ( defined( $data->{SSD_Drive_Prefix} ) ? 1 : 0 );
+    $text .=
+"# 12M NOTE: This is running with forced volumes - the size is irrelevant\n";
+    if ( defined( $data->{Drive_Prefix} ) ) {
```
dangogh: repeated code here..    should probably just loop ```foreach my $k (keys %$data) { ... }```

traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -749,38 +769,56 @@ sub hosting_dot_config {
...
+
+    my $server = $self->server_data($id);
+    my $text   = $self->header_comment( $server->host_name );
+    my $data   = $self->param_data( $server, $file );
+
+    my $has_multiple_volumes =
+      ( ( defined( $data->{Drive_Prefix} ) ? 1 : 0 ) +
+          ( defined( $data->{RAM_Drive_Prefix} ) ? 1 : 0 ) +
+          ( defined( $data->{SSD_Drive_Prefix} ) ? 1 : 0 ) ) > 1;
+    if ( defined( $data->{Drive_Prefix} ) ) {
```
dangogh: same comment here re repeated code..

traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -691,25 +691,45 @@ sub generic_config {
...
-	my $file = shift;
-
-	my $server = $self->server_data($id);
-	my $data   = $self->param_data( $server, "storage.config" );
-	my $text   = $self->header_comment( $server->host_name );
-	if ( defined( $data->{RAM_Drive_Prefix} ) ) {
+sub volume_dot_config_volume_text {
+    my $volume      = shift;
+    my $num_volumes = shift;
+    return
```
dangogh: more readable as ```"volume=$volume scheme=http size=$size%\n"```

robert-butts: That's how I had it before perltidy made it that way.

robert-butts: Ah, no it wasn't, heh. Single string. Done.

traffic_ops/app/lib/UI/ConfigFiles.pm
robert-butts: Hm, if we did that, it would add up other params someone made, e.g `Foo_Drive_Prefix`, or `RAN_Drive_Prefix`. 

We could `grep { /(|RAM|SSD)Drive_Prefix/ }`. That seems hard to read to me, but maybe I'm just not familiar enough with Perl.

PR #1081, opened 2016-02-20
Router and Monitor no longer require TRAFFIC_CONTROL_VERSION env var to run maven

### LINKED ISSUE #1080: Router and Monitor poms have version value that can't be referenced by external maven projects
The expression cannot be used by other java projects which we need to support immediately.

trevorackerman:
Fixes #1080 and should still support our build_rpm.sh


trevorackerman:
Dan, can you confirm that these changes indeed support the build scripts you wrote?

PR #1093, opened 2016-02-24
Add Traffic Ops 2.0 service to get ATS configs

robert-butts:
This service acts as a 'microservice' and serves ATS config files, by querying the Traffic Ops API.

This currently only creates storage.config, as a proof-of-concept. But it has the framework to easily add the rest of the configs. If we decided not to do the microservice thing, the config file logic easily be adapted to a library within a larger service. In fact, it would probably be good to make it a package anyway.


robert-butts:
To clarify, this pulls from the existing Traffic Ops 1.2 API. That's why it's still using the mojo cookie, etc. Once the 2.0 API is stable enough and has all the features this needs, I'll change this to use it. It should be pretty simple to adapt.

PR #1107, opened 2016-02-29
Test helper now deletes cachegroups wihout relying on order of ids

trevorackerman:


traffic_ops/app/lib/Test/TestHelper.pm
```diff
@@ -145,10 +145,21 @@ sub teardown {
 sub teardown_cachegroup {
 	my $self   = shift;
 	my $schema = shift;
-	my $cachegroups =
-		$schema->resultset("Cachegroup")->search( undef, { order_by => { -desc => [qw{parent_cachegroup_id secondary_parent_cachegroup_id}] } } );
-	while ( my $row = $cachegroups->next ) {
-		$row->delete();
+
+	my $cachegroups = $schema->resultset("Cachegroup");
+	while ( $cachegroups ne 0 ) {
```
dangogh: shouldn't need a while here,  and comparison isn't right on a couple of levels.  use != for numeric comparisons,  and ne for string.   $cachegroups is an object returned from resultset and not a number.   Did you want to use $cachegroups->count() == 0?

PR #1113, opened 2016-03-02
Make OPTIONS work with CORS, Rob probably has a fancier way, but this…

knutsel:
… works

### DIFF
--- a/traffic_ops/experimental/server/routes/routes.go
+++ b/traffic_ops/experimental/server/routes/routes.go
@@ -37,7 +37,7 @@ const apiPath = "/api/2.0/"
 // This function returns an http.Handler to be used in http.ListenAndServe().
 func CreateRouter(db *sqlx.DB) http.Handler {
 	router := mux.NewRouter().StrictSlash(true)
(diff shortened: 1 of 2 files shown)


traffic_ops/experimental/server/routes/routes.go
```diff
@@ -37,7 +37,7 @@ const apiPath = "/api/2.0/"
 // This function returns an http.Handler to be used in http.ListenAndServe().
 func CreateRouter(db *sqlx.DB) http.Handler {
 	router := mux.NewRouter().StrictSlash(true)
-	router.HandleFunc("/login", wrapHeaders(auth.GetLoginFunc(db), []api.ApiMethod{api.POST})).Methods("POST")
+	router.HandleFunc("/login", wrapHeaders(auth.GetLoginFunc(db), []api.ApiMethod{api.POST})).Methods("POST", "OPTIONS")
```
robert-butts: If we make the OPTIONS method a separate function, and register it separately, we can get rid of the `if` statement and push the logic into the router.

mitchell852: and while you're at it, if somebody can move /login to /api/2.0/login that would be very helpful.

PR #1117, opened 2016-03-02
Move all TO 2.0 API paths to /api/2.0, move conditional logic into the router

robert-butts:


knutsel:
I was going to say something on the ticket, sorry, too many things going on... Do we *really* want the /login to be under /api/2.0/? /login will give you a token for more than just /api/2.0. I don't know that that matters wanted to call it out. @mitchell852 ?

mitchell852:
I think from an api consumer perspective, it really makes sense to have all the api endpoints in one place no matter what they do. i mean, "login" is part of the api, right? and if so it belongs in the api namespace.
 
for example:

/api/2/login <-- this one authenticates me and gives me a token
/api/2/logout <-- this one expires my token
/api/2/foos <-- this one gives me foos
/api/2/bars <-- this one gives me bars

to me, they all do something...

and then when the version increases:

/api/3/login <-- maybe this didn't change so it points to a common login() method
/api/3/logout
/api/3/foos
/api/3/bars

PR #1124, opened 2016-03-03
Tm pom cleanup

trevorackerman:
In addition to removing unused items from the pom, now the fast running unit tests will run before the integration tests. See the Category annotation in MonitorApplicationTest and the maven-surefire-plugin and maven-failsafe-plugin configuration in the pom.

This does change how to run all the tests.

It used to be that running 'mvn test' would run all tests, now it runs only tests without Category annotation (meant to indicate unit tests)

To run all tests run either 'mvn integration-test' or 'mvn verify'

### DIFF
--- a/traffic_monitor/README.md
+++ b/traffic_monitor/README.md
@@ -1,5 +1,10 @@
 # Traffic Monitor
 
+### Manually building the rpm
+
+execute the following command:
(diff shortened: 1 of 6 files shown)


trevorackerman:
need to finish running tests

PR #1126, opened 2016-03-03
default of 0000-00-00 is not allowed in newer versions of mysql, prev…

knutsel:
…enting the migrations from working


knutsel:
actually, let me update this with the right seeds as well, don't merge yet.

knutsel:
ready to merge now. Thx.

PR #1145, opened 2016-03-09
always update IPv6 address and gateway

### LINKED ISSUE #1144: TO ipv6 configure for server is not removable
In Traffic Ops, IPv6 address for server can be added, modified. But remove IPv6 address will not take effect.
At the same time, when configure a IPv6 address for server without IPv6 gateway specified. An error message will promote that the IPv6 address (not the gateway) is invalid. The error message is quite confusing.

zhilhuan:
fix #1144

### DIFF
--- a/traffic_ops/app/lib/UI/Server.pm
+++ b/traffic_ops/app/lib/UI/Server.pm
@@ -372,8 +372,10 @@ sub check_server_input {
 		$err .= $paramHashRef->{'ip_address'} . " and " . $paramHashRef->{'ip_gateway'} . " are not in same network" . $sep;
 	}
 
-	if ( defined( $paramHashRef->{'ip6_address'} )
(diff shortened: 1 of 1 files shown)


traffic_ops/app/lib/UI/Server.pm
```diff
@@ -375,10 +375,13 @@ sub check_server_input {
 	if ( defined( $paramHashRef->{'ip6_address'} )
```
jheitz200: Can you please check if the user enters in an ip6_gateway.  Currently, if a user enters an ip6_gateway and no ip6_address the ui just silently ignores the value.

zhilhuan: @jheitz200 Yes, updated just now. Thanks for your comments.

PR #1149, opened 2016-03-11
Dockerfile fixes

robert-butts:
This adds Grafana and Heka to the Dockerfiles, and fixes the Traffic Vault Dockerfile to listen on 0.0.0.0.

### DIFF
--- a/docker/Traffic_Stats_Dockerfile_run.sh
+++ b/docker/Traffic_Stats_Dockerfile_run.sh
@@ -16,6 +16,7 @@
 start() {
 	service influxdb start
 	service traffic_stats start
+	service grafana-server start
 	touch /opt/traffic_stats/var/log/traffic_stats/traffic_stats.log
 	exec tail -f /opt/traffic_stats/var/log/traffic_stats/traffic_stats.log
 }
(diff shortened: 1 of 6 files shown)


docker/Traffic_Stats_Dockerfile_run.sh
```diff
@@ -69,10 +70,13 @@ init() {
 	influx -execute 'create retention policy daily on deliveryservice_stats duration 26h replication 3 DEFAULT'
 	influx -execute 'create retention policy daily_stats on daily_stats duration INF replication 3 DEFAULT'
```
dneuman64: The retention policy on daily_stats should be "indefinite" (I probably had you looking at out-of-date docs).  Also we want to create the monthly RPs for each database as well as the CQs.  Checkout what this is doing:  https://github.com/Comcast/traffic_control/blob/master/traffic_stats/influxdb_tools/create_ts_databases.go  I have an issue open for 1.5.0 to compile create_ts_databases and put it in the TS build, at that point all we need to do it run it.

dneuman64:
looks good to me, thanks @robert-butts!

PR #1169, opened 2016-03-16
Modified Traffic Ops to allow server types that begin with EDGE or MI…

elsloo:
…D such that we are not required to use specifically and only EDGE and MID as types for their respective tiers.


knutsel:
This looks good to me @elsloo - ran the tests and eye-balled the changes. Nice work.

dewrich:
@elsloo Make sure we don't forget about "custom" extensions ;)

PR #1172, opened 2016-03-16
Ort connection keepalive

mtorluemke:


traffic_ops/bin/traffic_ops_ort.pl
```diff
@@ -22,13 +22,15 @@
 use File::Path;
 use Fcntl qw(:flock);
 use MIME::Base64;
-use Data::Dumper;
+use LWP::ConnCache;
```
PSUdaemon: I don't think you need to use this directly. The docs suggest that you just set `keep_alive` to the `total_capacity` when calling `LWP::UserAgent->new()`. Also I'm not sure `undef` is a valid value.

traffic_ops/bin/traffic_ops_ort.pl
```diff
@@ -1040,53 +1042,57 @@ sub check_this_plugin {
...
-	if ( $url =~ m/\.json$/ ) {
-		eval {
-			decode_json($result);
-			1;
-		} or do {
-			my $error = $@;
-			( $log_level >> $FATAL ) && print "FATAL $url did not return valid JSON: $result | error: $error\n";
+
+	( $log_level >> $DEBUG ) && print "DEBUG Total connections in LWP cache: " . $lwp_conn->conn_cache->get_connections("https") . "\n";
+	my %headers = ( 'Cookie' => $cookie );
```
PSUdaemon: Why do we do this here *and* on line 548?

traffic_ops/bin/traffic_ops_ort.pl
```diff
@@ -22,13 +22,15 @@
 use File::Path;
 use Fcntl qw(:flock);
 use MIME::Base64;
-use Data::Dumper;
+use LWP::ConnCache;
+use LWP::UserAgent;
+use LWP::Protocol::https;
```
PSUdaemon: The docs suggest that you also do not use this directly.

traffic_ops/bin/traffic_ops_ort.pl
```diff
@@ -116,21 +118,23 @@
...
 
+#### LWP globals
+@LWP::Protocol::http::EXTRA_SOCK_OPTS = ( SendTE => 0, KeepAlive => 1, PeerHTTPVersion => "1.1" );
+my $lwp_conn                   = &setup_lwp(); 
+
 my $unixtime       = time();
 my $hostname_short = `/bin/hostname -s`;
 chomp($hostname_short);
 my $domainname = &set_domainname();
+$lwp_conn->agent("$hostname_short-$unixtime");
```
PSUdaemon: Can we add the Traffic Control version to this?

traffic_ops/bin/traffic_ops_ort.pl
```diff
@@ -116,21 +118,23 @@
 my $CFG_FILE_PREREQ_FAILED     = 3;
 my $CFG_FILE_ALREADY_PROCESSED = 4;
 
+#### LWP globals
+@LWP::Protocol::http::EXTRA_SOCK_OPTS = ( SendTE => 0, KeepAlive => 1, PeerHTTPVersion => "1.1" );
```
PSUdaemon: Do these apply if we are using the https protocol and not the http one?

traffic_ops/bin/traffic_ops_ort.pl
mtorluemke: Yeah, _if_ you set `keep_alive` when you create the `LWP::UserAgent` object, then you have to pass an initialized `LWP::ConnCache` object to the `conn_cache` method (really not any better or worse than the top-level include). However, `LWP::UserAgent` won't let you set `undef` for keep_alive, but `LWP::ConnCache` will for `total_capacity`. `undef` for the size of the cache is valid, though potentially not practical. 


See: https://api.metacpan.org/source/ETHER/libwww-perl-6.15/lib/LWP/ConnCache.pm, and specifically:

    ```if (defined(my $total = $self->{cc_limit_total})) {
	while (@$conns > $total) {
	    $self->dropping(shift(@$conns), "Total capacity exceeded");
	}
    }```

traffic_ops/bin/traffic_ops_ort.pl
mtorluemke: Agree, I think this can be cleaned up.

traffic_ops/bin/traffic_ops_ort.pl
mtorluemke: Yes, but might be a less confusing way to set them.

traffic_ops/bin/traffic_ops_ort.pl
mtorluemke: Good idea. https://github.com/Comcast/traffic_control/issues/1174

traffic_ops/bin/traffic_ops_ort.pl
mtorluemke: 548 is a POST (to clear the update status in Traffic Ops). Line 1045 starts the sub that only does GETs. Likely would be more overall code to make one sub that does both, or one sub for each method. LWP also has a cookie jar that is likely the best solution.

traffic_ops/bin/traffic_ops_ort.pl
PSUdaemon: Mmm, exists vs defined. Ok, fair enough. Seems like an oversight on *their* part, but perhaps not. Is there a reason that we need more than 1 connection?

mtorluemke: In the future, I can see us load-balancing or failing over across different Traffic Ops endpoints. Really, I just didn't want to pick a value. Thoughts on initial value?

PSUdaemon: I was thinking about this and something like 100 seemed fine. How many different GET's do we do?

But after that, I was thinking of the ways we'd use more than 1. I haven't looked at the code in depth, but it looks like it uses hostname as a key. So I suspect it won't resolve DNS again. And if we used a load balancer we'd still terminate the TCP connection to some host. And if there was a failure I suspect it'd reconnect anyway, or we should in the even of a connection error.

If we used it in a multi-threaded or event driven way I could see us doing multiple connections, but at that point maybe we'd rewrite it in Go or be using Puppet or something else entirely.

If I haven't convinced you otherwise, I'm fine with staying with the code as is. Just wanted to talk this detail out.

mtorluemke: Today, between 60 and 70, depending on the config. 100 is likely more than safe, in any situation.

Issue #1156, opened 2016-03-14
Traffic Ops Delivery Service change for Max DNS requests returns to 0


G4m30v3r:
Editing Max DNS requests changes it to default 0 on save.

elsloo:
@G4m30v3r I just tested this in master (what will become 1.5.0) and I cannot reproduce the issue.

G4m30v3r:
produced this consistently on 1.4.0.2426. Showed output to elsloo directly.

elsloo:
Confirmed that this affects 1.4.x and a fix needs to be applied; most likely a backport from a commit that has already been made to master.

dneuman64:
closed by #1182

PR #1182, opened 2016-03-17
fix deliveryservice type references

dangogh:
This should have been cherry-picked earlier.   Resolves the "Odd number of elements..." error message and not saving updated deliveryservice.


dangogh:
Fixes #1156

PR #1188, opened 2016-03-18
Add support for providing geolocation coordinates on a coverage zone entry to improve routing accuracy

elsloo:


traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/NetworkNode.java
```diff
@@ -120,11 +131,16 @@ public static NetworkNode generateTree(final JSONObject json) {
     }
 
     public NetworkNode(final String str) throws NetworkNodeException {
-        this(str, null);
+        this(str, null, null);
```
trevorackerman: Don't really need this change

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -462,29 +462,33 @@ public HTTPRouteResult route(final HTTPRequest request, final Track track) throw
 		return routeResult;
 	}
 
-	protected CacheLocation getCoverageZoneCache(final String ip) {
+	@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.NPathComplexity"})
```
trevorackerman: I'd recommend finding a way to do these changes so the suppress warnings aren't needed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -492,6 +496,20 @@ protected CacheLocation getCoverageZoneCache(final String ip) {
 			}
 		}
 
+		/*
+		 * We had a hit in the CZF but the name does not match a known cache location.
+		 * Check whether the CZF entry has a geolocation and use it if so.
+		 */
+		if (nn.getGeolocation() != null) {
```
trevorackerman: writing this as the following might eliminate the need for suppressing warnings from good old PMD

`if (nn.getGeolocation == null) {
   return null
}

final List<CacheLocation> cacheLocations = ...
...`

If that doesn't work it might be worthwhile to put the lookup of the cacheLocation in its own small method.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/NetworkNode.java
elsloo: Technically true, however when written this way we have one less constructor to pass through to reach the one in the above change. That said, I don't feel strongly about it.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
elsloo: The above suggestion wasn't enough so I refactored a little more which allowed me to remove the SuppressWarnings annotation.

PR #1190, opened 2016-03-21
Add TO 2.0 Relational Caching Proxy config

robert-butts:
Starting point for relational, typed ATS parameters.

Adds the ATS config SQL tables, Go structs and API endpoint, and a conversion script to convert old parameters from 1.x.

This adds considerable data safety. For example, it isn't possible to add a `proxy.config.cluster.cluster_port` value of `foo`, or anything that isn't a valid port. Currently, the only constraints are for ports and ATS enum values. But Postgres constraints are very powerful, including regex. For example, we can make a regex constraint for URI fields.

Again, this is only the first step. The next steps that need done are:

* adding all ATS config files and parameters in SQL and REST (currently it's mostly only record.config)

* an ORT-like script to run on the ATS client, and fetch the endpoint and transform it into ATS config files

* Web GUI components for viewing and editing

* A script to add a new column and re-generate the Go endpoint code
  * This will be less necessary if/when we add all ATS config options to the config files. But until then, it should be easy to add new config options when they're needed.


knutsel:
I think we should think about this a bit more; biggest problem i have with this now, is the large amount of ATS specific things that are going into the generic schema....

PR #1205, opened 2016-03-22
Fix postinstall prompting for TO user every time

### LINKED ISSUE #1204: Postinstall prompts for admin user and pass on every run
Postinstall shouldn't ask for the admin credentials on every update run. It should only need that info the first time.

The reason it needs it every time now, is to check if default profiles exist, to create them if they don't.

The long-term solution is to make postinstall store user-entered data in a config file, so when it's run it checks the config first and doesn't re-prompt for things the user has already entered. We don't want to store the password for security reasons, but we can still fix this issue by storing whether defaults have been created.

The short-term solution is to write an empty file (`.reconfigure_defaults`) when defaults have been created, and not even prompt for user+pass to check, if that file exists.

robert-butts:
Fixes #1204

### DIFF
--- a/traffic_ops/install/bin/postinstall
+++ b/traffic_ops/install/bin/postinstall
@@ -42,6 +42,7 @@ my $profile_dir       = "/opt/traffic_ops/install/data/profiles/";
 my %dbdriver          = ( mysql => "mymysql", );
 
 my $reconfigure = "/opt/traffic_ops/.reconfigure";
+my $reconfigure_defaults = "/opt/traffic_ops/.reconfigure_defaults";
 my $tmAdminUser = "";
 my $tmAdminPw = "";
 my $parameters;
@@ -612,6 +613,10 @@ print "\nWaiting for Traffic Ops to start.\n\n";
 sleep(5);
 
 sub profiles_exist {
+		if ( -f $reconfigure_defaults ) {
+				return 1;
+		}
+
 		while ( length $tmAdminUser == 0 ) {
 				$tmAdminUser =
 						InstallUtils::promptUser("Administration username for Traffic Ops");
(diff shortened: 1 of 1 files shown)


traffic_ops/install/bin/postinstall
```diff
@@ -667,6 +672,7 @@ sub profiles_exist {
 		foreach my $profile (@$profiles_response) {
 				if ( !exists $initial_profiles{ $profile->{"name"} } ) {
 						print "Found existing profile (" . $profile->{"name"} . ")\n";
+						open( my $reconfigure_defaults_file, '>>', $reconfigure_defaults ) or die("Failed to open() $reconfigure_defaults: $!");
```
dangogh: why create it both here and line 687?   I think this one should go away...

traffic_ops/install/bin/postinstall
```diff
@@ -677,6 +683,9 @@ if ( !profiles_exist() ) {
 		print "Creating default profiles...\n";
 		replace_profile_templates();
 		import_profiles();
+		if ( profiles_exist() ) {
+				open( my $reconfigure_defaults_file, '>>', $reconfigure_defaults ) or die("Failed to open() $reconfigure_defaults: $!");
```
dangogh: should not be appending to the file -- single '>' should be good.

and should close($reconfigure_defaults_file) as well

robert-butts: If we open for overwrite instead of append, won't it delete anything in the file? This way, if someone puts something in that file for whatever reason, it won't destroy their data.

I'll fix it to close.

traffic_ops/install/bin/postinstall
robert-butts: If this one goes away, the other one will never get called. We could remove the second open, because the profiles_exist() call will implicitly create it; but we'll still have to call profiles_exist() the second time, because the file doesn't get created on the first call. So
```if ( profiles_exist() ) {
				open( my $reconfigure_defaults_file, '>>', $reconfigure_defaults ) or die("Failed to open() $reconfigure_defaults: $!");
		}```

could be changed to just `profiles_exist()`.

dangogh: either way..   I think there should be just one..

traffic_ops/install/bin/postinstall
dangogh: they shouldn't put anything in there -- my comment earlier was *we* could put something in there as documentation to indicate why the file exists, but they should never count on anything they put in there to be safe..

traffic_ops/install/bin/postinstall
robert-butts: Ok, removed the other one.

PR #1209, opened 2016-03-23
Add Traffic Ops secure parameter

jifyang:
Can fix the issue #867: "Admin password visible to Operator User in TrafficOps UI".

The design document:
    https://docs.google.com/document/d/11urRVIs2qyTidAAUn-kYcHCPuLnMPZFmP5qrvDQYJZE/edit?usp=sharing


mitchell852:
Can you run `db/admin.pl reverse_schema` to update the dbix files?

jifyang:
The dbix file is updated. It is the file app/lib/Schema/Result/Parameter.pm.
I tested: after update to the newest code, this file should be same as the committed file in this pull request.

dewrich:
@jifyang I verified that everything looks good for this feature, the only thing I had to fix is rename the migration because it's now out of sequence.  Please rename it to today's date and I'll merge this through

dewrich:
@jifyang I also just noticed that there are no test cases, can you please add them?

jifyang:
Updated for DB migration file name modification and test case.

PR #1212, opened 2016-03-23
adds geolocation provider per delivery service

### LINKED ISSUE #1202: TO: Add per-delivery service geo provider
Currently, the default is assumed to be MaxMind. Moving forward, we want to at least make Neustar available as a configuration option on a delivery service. 

Should we pop up a notification to inform the user of where to go to get a Neustar license? Not sure...I think we're okay without it.

dangogh:
MaxMind is default,   Neustar is an option.

Fixes #1202


dangogh:
Hmm..   looking back at it,   not sure I covered the json api endpoints..

traffic_ops/app/lib/Fixtures/Deliveryservice.pm
```diff
@@ -22,138 +23,141 @@ my %definition_for = (
...
+			miss_long             => '-87.627778',
+			check_path            => '/crossdomain.xml',
+			type                  => 8,
+			profile               => 3,
+			cdn_id                => 1,
+			ipv6_routing_enabled  => 1,
+			protocol              => 1,
+			display_name          => 'test-ds1-displayname',
+			initial_dispersion    => 1,
 			regional_geo_blocking => 1,
```
jheitz200: should geo_provider be included in this fixture?

dangogh: I was leaving it out of that one to make sure default of Maxmind (0) works

PR #1214, opened 2016-03-24
Fix TO remap rules for nonstandard ports

### LINKED ISSUE #1134: TO: non-standard ports need added to remap rules
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

A "host" without any trailing port information implies the default port for the service requested (e.g., "80" for an HTTP URL).

robert-butts:
Fixes Traffic Ops remap.config generation to add the port to HTTP
delivery service remap rules, if it's nonstandard. This fixes the Host
headers generated by ATS to conform to RFC 2616, and fixes the CDN to
work with edges with nonstandard ports.

Fixes #1134

### DIFF
--- a/traffic_ops/app/lib/UI/ConfigFiles.pm
+++ b/traffic_ops/app/lib/UI/ConfigFiles.pm
@@ -230,7 +230,11 @@ sub ds_data {
 				$re =~ s/\\//g;
 				$re =~ s/\.\*//g;
 				my $hname = $ds_type =~ /^DNS/ ? "edge" : "ccr";
-				my $map_from = "http://" . $hname . $re . $ds_domain . "/";
+				my $portstr = "";
+				if ( $hname eq "ccr" && $server->tcp_port > 0 && $server->tcp_port != 80 ) {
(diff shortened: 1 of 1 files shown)


traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -230,7 +230,11 @@ sub ds_data {
 				$re =~ s/\\//g;
 				$re =~ s/\.\*//g;
 				my $hname = $ds_type =~ /^DNS/ ? "edge" : "ccr";
-				my $map_from = "http://" . $hname . $re . $ds_domain . "/";
+				my $portstr = "";
+				if ( $hname eq "ccr" && $server->tcp_port ne "" && $server->tcp_port ne "80" ) {
```
mtorluemke: Does $server->tcp_port come through as a string? I thought it would be an int, and thus `== 80` would be needed here...

robert-butts: Yes, it's a number. It worked because Perl cast the number to a string. I'll fix it.

PR #1224, opened 2016-03-25
Add TO2.0 Snapshot CRConfig

robert-butts:
We discussed making the CRConfig something more complex like a state machine. But that's a big feature. This was quick to write, and gives us the existing 1.x functionality, until we have time for the larger feature.


knutsel:
A bit worried about where the DB is now... Some of these changes xmlId -> Name in DS for example seem outside of the scope? Merging anyway, I'm sure you know what you're doing, and is is called 'experimental'.

robert-butts:
They are out of scope. The current schema doesn't work with the code. I didn't realise it until halfway thru this feature, so I made the changes necessary to make it work. I'll fix the code to fully match the schema in another PR.

PR #1233, opened 2016-03-30
Fix TO Clone DS Assignments to exclude other CDNs

### LINKED ISSUE #677: Cloning DS assignments can list caches of a different CDN
when cloning DS assignments, any server can be selected even if it's a member of the opposite type.
ie. I can assign a a profile association to type A and B when my server profile is a type A

robert-butts:
Fixes #677

### DIFF
--- a/traffic_ops/app/lib/UI/DeliveryServiceServer.pm
+++ b/traffic_ops/app/lib/UI/DeliveryServiceServer.pm
@@ -31,18 +31,6 @@ sub cpdss_iframe {
 	if ( $mode eq "view" ) {
 		my $server = $self->db->resultset('Server')->search( { 'me.id' => $srvr_id } )->single();
 
-		my $valid_profiles;
-		my $psas = $self->db->resultset('Server')->search(
-			{ cdn_id => $server->cdn_id },
-			{
-				select   => 'profile',
-				distinct => 1
-			}
-		)->get_column('profile');
-		while ( my $row = $psas->next ) {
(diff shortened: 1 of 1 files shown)


traffic_ops/app/lib/UI/DeliveryServiceServer.pm
```diff
@@ -44,7 +44,7 @@ sub cpdss_iframe {
 		}
 
 		my @etypeids = &type_ids( $self, 'EDGE%', 'server' );
-		my $rs = $self->db->resultset('Server')->search( { type => { -in => \@etypeids } }, { prefetch => 'profile', order_by => 'host_name' } );
+		my $rs = $self->db->resultset('Server')->search( { type => { -in => \@etypeids }, cdn_id => $server->cdn_id }, { prefetch => 'profile', order_by => 'host_name' } );
```
mtorluemke: I think if we do this, we can get rid of the profile check -- lines 35-40 gets a list of valid profiles, and that includes a check for the same CDN.

robert-butts:
Also, if it isn't immediately obvious (it wasn't to me) why the prior profile check (which I just removed in the 2nd commit), which filters other CDNs, wasn't excluding servers on other CDNs:

It's because the profile check excludes profiles on servers on other CDNs, but then includes profiles on the current server's CDN regardless. So, if there's a server on a different CDN that shares a profile with one on the current server's CDN, those servers on that "multi-CDN profile" wouldn't be filtered out.

PR #1239, opened 2016-03-30
Fix TO to not add remap rules for Live DSs on mids

### LINKED ISSUE #752: Remap rules shouldn't be generated for MID caches for DNS_LIVE delivery services
DNS_LIVE DS's don't route traffic through the mids, so we shouldn't generate those remap rules. HTTP_LIVE (and others?) might have the same problem.

robert-butts:
Fixes #752

Removing remap rules makes me anxious, so, just to confirm this is the right thing to do: this removes _**all**_ rules in `remap.config` on Mids for Delivery Services which are LIVE and not NATL.

### DIFF
--- a/traffic_ops/app/lib/UI/ConfigFiles.pm
+++ b/traffic_ops/app/lib/UI/ConfigFiles.pm
@@ -926,7 +926,9 @@ sub remap_dot_config {
 	if ( $server->type->name =~ m/^MID/ ) {
 		my %mid_remap;
 		foreach my $remap ( @{ $data->{dslist} } ) {
-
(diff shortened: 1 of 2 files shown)


mtorluemke:
Does this play nice with the other fields that cause the generation of the remap.config line anyway? Meaning -- is a 'location' parameter created for the MID profiles, and are contents of those files populated? (hdr_rw_<xml_id>.config is probably the best example)

robert-butts:
It was still creating the location parameters. Fixed with 1b0a76, which should also fix the ORT endpoint to not include the `hdr_rw_` paths. I checked the other files, `regex_remap` and `cacheurl` and neither appear to have live local deliveryservice rules.

Issue #1253, opened 2016-04-01
Traffic Ops should always assign a volume to each drive in ATS's storage.config


jschmerge:
When replacing hard drives, ATS has an issue picking up the replaced hard drives unless they have an explicit volume assigned to them in storage.config

By always adding a volume=x parameter to the lines in storage.config, this issue is side-stepped.

elsloo:
This was resolved with #1255.

PR #1255, opened 2016-04-01
Always insert a 'volume=x' parameter into storage.config

jschmerge:
ATS has an issue where it cannot figure out what volume to assign
to a replacement drive if other drives already have an existing
volume assignment. Explicitly setting the volume parameter for every
drive in storage.config works around this issue.

This is a fix for issue #1253

### DIFF
--- a/traffic_ops/app/lib/UI/ConfigFiles.pm
+++ b/traffic_ops/app/lib/UI/ConfigFiles.pm
@@ -815,15 +815,12 @@ sub storage_dot_config_volume_text {
     my $prefix               = shift;
     my $letters              = shift;
     my $volume               = shift;
-    my $has_multiple_volumes = shift;
 
     my $text = "";
     my @postfix = split( /,/, $letters );
     foreach my $l ( sort @postfix ) {
         $text .= $prefix . $l;
-        if ($has_multiple_volumes) {
-            $text .= " volume=" . $volume;
-        }
+        $text .= " volume=" . $volume;
         $text .= "\n";
     }
     return $text;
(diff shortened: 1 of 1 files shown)


traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -838,24 +835,34 @@ sub storage_dot_config {
...
         );
     }
     if ( defined( $data->{RAM_Drive_Prefix} ) ) {
+        if ( defined($data->{RAM_Volume} ) ) {
+            $assigned_volume = $data->{RAM_Volume};
+        }
         $text .= storage_dot_config_volume_text(
             $data->{RAM_Drive_Prefix}, $data->{RAM_Drive_Letters},
-            $data->{RAM_Volume},       $has_multiple_volumes
+            $data->{RAM_Volume}
```
robert-butts: Does this need to pass `$assigned_volume`?

traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -838,24 +835,34 @@ sub storage_dot_config {
...
         );
     }
     if ( defined( $data->{SSD_Drive_Prefix} ) ) {
+        if ( defined($data->{SSD_Volume} ) ) {
+            $assigned_volume = $data->{SSD_Volume};
+        }
         $text .= storage_dot_config_volume_text(
             $data->{SSD_Drive_Prefix}, $data->{SSD_Drive_Letters},
-            $data->{SSD_Volume},       $has_multiple_volumes
+            $data->{SSD_Volume}
```
robert-butts: Does this need to pass `$assigned_volume`?

traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -838,24 +835,34 @@ sub storage_dot_config {
...
 
     if ( defined( $data->{Drive_Prefix} ) ) {
+        if ( defined($data->{Disk_Volume} ) ) {
+            $assigned_volume = $data->{Disk_Volume};
+        }
         $text .= storage_dot_config_volume_text(
             $data->{Drive_Prefix}, $data->{Drive_Letters},
-            $data->{Disk_Volume},  $has_multiple_volumes
+            $assigned_volume
         );
```
robert-butts: Should we increment `$assigned_volume` here, and after we call `storage_dot_config_volume_text` below as well? If we don't, and `$data->{RAM_Volume}` or `$data->{SSD_Volume}` aren't set but their `_Drive_Prefix` is, they'll wrongly get the previously set volume.

traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -838,24 +835,42 @@ sub storage_dot_config {
...
     }
+
+	# Default to 2 here
+	$assigned_volume = 2;
```
elsloo: What happens if you have a RAM_Drive_Prefix parameter in the profile but no RAM_Volume parameter, and the Disk_Volume parameter is set to 2?

jschmerge: Shouldn't the UI prevend that?

elsloo: Should and does are two separate things. These particular configuration items are just parameters in the parameter table that are mapped to a profile. That means there's zero field validation happening; the wild wild west. Anything is possible; they could set RAM_Volume to X if they wanted to.

PR #1260, opened 2016-04-01
ORT: handle case where plugin has no config file. fixes #1259.

### LINKED ISSUE #1259: ORT: handle case where remap.config plugin has no config file

mtorluemke:


mtorluemke:
Spoke with @smalenfant -- merging for 1.5 -- will revisit later if needed.

PR #1263, opened 2016-04-04
To2 fix code for schema

robert-butts:
Like the commit message says, there's still a lot of work that needs done. I'll be working on all the things listed. If we don't feel comfortable merging as-is, I can close this and make another PR when everything is done. But this makes the code compile with the current `schema.sql`, which is more than the current code does.


robert-butts:
I don't recommend reviewing the `api/` files, except `api.go`. They are all generated, with no modifications.

robert-butts:
Note this depends on the separate PR #1227.

PR #1277, opened 2016-04-06
Updated Traffic Stats to handle DS Type data

dneuman64:
Also updated the DS scripted dashboard to account for type data.

### DIFF
--- a/traffic_stats/grafana/traffic_ops_deliveryservice.js
+++ b/traffic_stats/grafana/traffic_ops_deliveryservice.js
@@ -275,7 +275,7 @@ dashboard.refresh = "30s";
           "steppedLine": false,
           "targets": [
             {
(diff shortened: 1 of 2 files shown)


traffic_stats/grafana/traffic_ops_deliveryservice.js
```diff
@@ -275,7 +275,7 @@ dashboard.refresh = "30s";
           "steppedLine": false,
           "targets": [
             {
-              "query": "SELECT mean(value)*1000 FROM \"monthly\".\"kbps.cg.1min\" WHERE deliveryservice='" + which + "' and cachegroup != 'total' and $timeFilter GROUP BY time(60s), cachegroup",
```
dneuman64: fyi...the 'total' cachegroup is already filtered out by the CQ, so this really wasn't necessary to being with.

PR #1281, opened 2016-04-06
uniquify parent info for parent.config

### LINKED ISSUE #1279: TO: parent.config in multi-site origin shows duplicate entries on each line
Sometimes, you get parent.config like this:

dest_domain=origin1.origin.com parent="1.2.3.4:80|1.0;1.2.3.4:80|1.0;" round_robin=consistent_hash go_direct=false parent_is_proxy=false

dangogh:
Fixes #1279

### DIFF
--- a/traffic_ops/app/lib/UI/ConfigFiles.pm
+++ b/traffic_ops/app/lib/UI/ConfigFiles.pm
@@ -1103,6 +1103,8 @@ sub parent_dot_config {
 				foreach my $parent ( @{ $pinfo->{$org_fqdn} } ) {
 					push @parent_info, format_parent_info($parent);
 				}
+				my %seen;
+				@parent_info = grep { !$seen{$_}++ } @parent_info;
 
 				my $parents = 'parent="' . join( '', @parent_info ) . '"';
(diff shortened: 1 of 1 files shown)


traffic_ops/app/lib/UI/ConfigFiles.pm
```diff
@@ -1147,9 +1149,13 @@ sub parent_dot_config {
 						push @secondary_parent_info, $ptxt;
 					}
 				}
+				my %seen;
+				@parent_info = grep { !$seen{$_}++ } @parent_info;
 				my $parents = 'parent="' . join( '', @parent_info ) . '"';
 				my $secparents = '';
 				if ( scalar @secondary_parent_info > 0 ) {
+					my %seen;
```
elsloo: Will redeclaring %seen here cause problems with this section of logic?

dangogh: no -- it's declaring it locally in that scope.  If there are no secondary parents,  it'll never get allocated.  If there are, this %seen goes away when it exits the block at line 1160

elsloo: Makes sense; thanks.

PR #1284, opened 2016-04-07
Traffic Router now runs as one application

trevorackerman:
Instead of running a separate application for the API, now use 2 connectors on one servlet.
Instead of using the TRServlet class, pass everything through a servlet filter for everything coming in on the main port (80). The filter does the same exact routing responses as the TRServlet class, in fact hardly any code has changed here other than a class name change and implementing the filter interface.

The api maven module's code is now inside the core maven module
The testServer module isn't needed anymore, it's gone too

The TrafficRouterStart class now makes use of a class named CatalinaTrafficRouter, this runs an 'embedded' tomcat container that uses the server.xml in core/src/main/opt/tomcat/conf

This will make it easy to test traffic router without installing the tomcat application and should keep it consistent with the currently supported version of Tomcat (6.0.33)

A test suite was added for blackbox integration testing against the above embedded traffic router.
Simple tests are there for a starting point. All these tests use an http client to send requests to a traffic router running in a separate process.

They're different from the existing integration tests which are whitebox tests, i.e. they dig directly into the guts of a running traffic router while they themselves are running inside the same java process.


traffic_router/core/src/test/java/com/comcast/cdn/traffic_control/traffic_router/core/integration/RouterTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+	@Test
+	public void itRedirectsValidRequests() throws IOException, InterruptedException {
+		// Traffic Router will give us a 503 until it is ready to route
+		// It also gives us a 503 when we don't make a valid routing request
+		// The following request though *SHOULD* work so try and do this request multiple times
+		// until we get a 302 to determine that all the application context is finished before
+		// starting tests
+
+		HttpGet httpGet = new HttpGet("http://localhost:8888/stuff?fakeClientIpAddress=113.203.235.227");
```
dneuman64: can you clean up this IP?

traffic_router/core/src/test/java/com/comcast/cdn/traffic_control/traffic_router/core/integration/RouterTest.java
```diff
@@ -0,0 +1,128 @@
...
+				triesLeft = 0;
+			} finally {
+				if (response != null) response.close();
+			}
+		}
+	}
+
+	@Test
+	public void itRejectsInvalidRequests() throws IOException {
+		HttpGet httpGet = new HttpGet("http://localhost:8888/stuff?fakeClientIpAddress=113.203.235.227");
```
dneuman64: can you clean up this IP?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/http/RouterFilter.java
```diff
@@ -16,76 +16,67 @@
...
 	@Override
-	public void init(final ServletConfig config) throws ServletException {
+	public void doFilterInternal(final HttpServletRequest request, final HttpServletResponse response, final FilterChain chain) throws IOException, ServletException {
+		final Date requestDate = new Date();
 
-		final ApplicationContext context = WebApplicationContextUtils.getWebApplicationContext(config
-				.getServletContext());
-		setTrafficRouterManager(context.getBean(TrafficRouterManager.class));
-		setStatTracker(context.getBean(StatTracker.class));
+		if (request.getLocalPort() == 3333) {
```
trevorackerman: Need to change this so it aligns with data coming from cr-config

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
```diff
@@ -89,63 +89,69 @@ public void init() {
 		scheduledService = executorService.scheduleWithFixedDelay(updater, pollingInterval, pollingInterval, TimeUnit.MILLISECONDS);
 	}
 
-	@SuppressWarnings("PMD.CyclomaticComplexity")
+	@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.NPathComplexity"})
 	public boolean updateDatabase() {
 		if (!databasesDirectory.exists() && !databasesDirectory.mkdirs()) {
 			LOGGER.error(databasesDirectory.getAbsolutePath() + " does not exist and cannot be created!");
 		}
```
dneuman64: should return false here

traffic_router/core/src/main/webapp/WEB-INF/web.xml
```diff
@@ -27,46 +27,32 @@
...
-        <url-pattern>/route/*</url-pattern>
-    </servlet-mapping>
-    <servlet-mapping>
-        <servlet-name>traffic_router</servlet-name>
         <url-pattern>/*</url-pattern>
     </servlet-mapping>
 
+    <servlet-mapping>
+        <servlet-name>status</servlet-name>
+        <url-pattern>/crs/*</url-pattern>
```
trevorackerman: Test that http://<fqdn>/crs/blah still goes through router filter (i.e. normal routing port)

traffic_router/core/src/test/java/com/comcast/cdn/traffic_control/traffic_router/core/CatalinaTrafficRouter.java
```diff
@@ -0,0 +1,83 @@
...
+		catalina.process(new String[] {"-config", serverXmlPath});
+		catalina.load();
+
+		// Override the port and app base property of server.xml
+		StandardService trafficRouterService = (StandardService) catalina.getServer().findService("traffic_router_core");
+
+		Connector[] connectors =trafficRouterService.findConnectors();
+		for (Connector connector : connectors) {
+			if (connector.getPort() == 80) {
+				connector.setPort(8888);
```
trevorackerman: log it to stdout

PR #1289, opened 2016-04-11
WIP: Mysql to postgres migration

dangogh:
(WIP) This is a Work In Progress -- @robert-butts please look over,  esp the region_pkey migration (last file).    This is the pattern for changing the id key to a natural key along with updating where it's used as a foreign key.


traffic_ops/experimental/server/db/convert.sh
```diff
@@ -0,0 +1,31 @@
...
+   modify column address varchar(128) null,
+   modify column city varchar(128) null,
+   modify column state varchar(2) null,
+   modify column zip varchar(5) null;
+alter table regex
+   modify column pattern varchar(255) null;
+alter table to_extension
+   modify column version varchar(45) null,
+   modify column info_url varchar(45) null,
+   modify column script_file varchar(45) null;
```
robert-butts: Why are these being made nullable?

traffic_ops/experimental/server/db/migrations/0002_cleanup.sql
```diff
@@ -0,0 +1,29 @@
...
+ALTER TABLE job_result
+	DROP CONSTRAINT job_result_agent_fkey,
+	DROP CONSTRAINT job_result_job_fkey;
+
+DROP TABLE hwinfo;
+DROP TABLE job;
+DROP TABLE job_agent;
+DROP TABLE job_status;
+DROP TABLE job_result;
+DROP TABLE servercheck;
```
robert-butts: Will the data in these tables be kept and converted somewhere before they're dropped? Do we need to?

traffic_ops/experimental/server/db/migrations/0001_initial_conversion_from_mysql.sql
```diff
@@ -0,0 +1,1235 @@
...
+SET default_with_oids = false;
+
+CREATE TABLE asn (
+    id integer NOT NULL,
+    asn integer NOT NULL,
+    cachegroup integer DEFAULT 0 NOT NULL,
+    last_updated timestamp without time zone DEFAULT now()
+);
+
+ALTER TABLE asn OWNER TO touser;
```
robert-butts: Hm, people can set the MySQL user to something other than 'touser' in postinstall. Does Goose have a way to use a variable here? Or is it possible to query the right user from SQL somehow?

traffic_ops/experimental/server/db/migrations/0001_initial_conversion_from_mysql.sql
```diff
@@ -0,0 +1,1235 @@
...
+    NO MAXVALUE
+    CACHE 1;
+
+ALTER TABLE asn_id_seq OWNER TO touser;
+
+ALTER SEQUENCE asn_id_seq OWNED BY asn.id;
+
+CREATE TABLE cachegroup (
+    id integer NOT NULL,
+    name character varying(45) NOT NULL,
```
robert-butts: We should consider making all text columns 'text' type. Postgres is fast and our data is small, we're almost certainly not going to have performance issues, and it would prevent bugs like #1302.

If we did have a performance issue in the future, we could always make affected columns varchar(x) then.

PSUdaemon: http://www.postgresql.org/docs/current/static/datatype-character.html

```
Tip: There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs and slower sorting. In most situations text or character varying should be used instead.
```

Unless you have a legitimate constraint to the length of a text field, use `text` instead of `varchar`.

PSUdaemon: @robert-butts Also, just using Postgres will prevent bugs like #1302 because it will not silently alter your data for you.

traffic_ops/experimental/server/db/convert.sh
dangogh: Using AnatolyUss/nmig for conversion -- these were to work around this issue:
https://github.com/AnatolyUss/nmig/issues/17

traffic_ops/experimental/server/db/migrations/0002_cleanup.sql
dangogh: dunno...   going off the current schema.sql -- we'll adjust as necessary

traffic_ops/experimental/server/db/migrations/0001_initial_conversion_from_mysql.sql
dangogh: no argument here..   changing to text type should make things much cleaner...

dangogh:
closing this -- postgres work is going on in another PR

PR #1307, opened 2016-04-14
get snapshot_dir param once; get reported servers as part of initial …

### LINKED ISSUE #1304: Non-performant code causing purge (invalidate content) api to timeout
POST /api/1.2/user/current/jobs creates a purge job which involves inserting a record in the Job table, snapshotting regex_revalidate.config and setting upd_pending flag on each server of the applicable CDN to 1. 

This process takes too long and causes a timeout.

I suspect the offending code is in snapshot_regex_revalidate() of https://github.com/Comcast/traffic_control/blob/master/traffic_ops/app/lib/MojoPlugins/Job.pm

specifically, the identical query performed for each server found on line 58 and 59.

dangogh:
fixes #1304 -- initial query of servers now only retrieves REPORTED;  config_file param retrieved only once instead of once per server.

@mitchell852 please review and help verify this change...


traffic_ops/app/lib/MojoPlugins/Job.pm
```diff
@@ -34,34 +36,42 @@ sub register {
...
+			my $m_host   = $self->req->url->base->host;
+			my $m_port   = $self->req->url->base->port;
 			my %cdn_domain;
+			my $snapshot_rs =
+				$self->db->resultset('Parameter')->search( { name => "snapshot_dir" }, { config_file => REGEX_CONFIG } )->single();
+			if (!defined $snapshot_rs) {
+				Mojo::Exception->throw("snapshot_dir parameter for config_file " . REGEX_CONFIG . " not found");
+				return;
+			}
+	  		my $snapshot_dir = $snapshot_rs->value;
```
mitchell852: could you put a comment about what exactly is going on in this while loop? i think i know but i'm not 100% sure.

traffic_ops/app/lib/MojoPlugins/Job.pm
```diff
@@ -34,34 +36,43 @@ sub register {
...
 			my %cdn_domain;
+			my $snapshot_rs =
+				$self->db->resultset('Parameter')->search( { name => "snapshot_dir" }, { config_file => REGEX_CONFIG } )->single();
+			if ( !defined $snapshot_rs ) {
+				Mojo::Exception->throw( "snapshot_dir parameter for config_file " . REGEX_CONFIG . " not found" );
+				return;
+			}
+			my $snapshot_dir = $snapshot_rs->value;
 
+			# generate regex_revalidate.config for one server for each cdn
```
mitchell852: this is incorrect behavior. it should ONLY generate regex_revalidate.config for the CDN that the delivery service belongs to...

dangogh:
@mitchell852 : updated to reduce number of db calls -- only need to generate for cdn for given ds.  Ready again for test..

PR #1310, opened 2016-04-18
add restful API: clone_siblings_ds

jifyang:
The APIs document is at https://docs.google.com/document/d/13GjIizusc7iUZq7TuVrsp8JZ2eEx52xUtoX9uRNzFgc/edit?usp=sharing

### DIFF
--- a/docs/source/development/traffic_ops_api/v12/server.rst
+++ b/docs/source/development/traffic_ops_api/v12/server.rst
@@ -780,3 +780,40 @@ Server
   +------------------+--------+------------------------------------------------+
   
 |
+
+**POST /api/1.2/servers/{:id}/clone_siblings_ds**
+
+  Assign delivery service to the server with the same as sibling server in the cache group.
+
+  Authentication Required: Yes
+
+  Role(s) Required: admin or oper
+
+  **Request Route Parameters**
+
+  +------------+----------+--------------------------------------------------------+
+  |    Name    | Required |           Description                                  |
+  +============+==========+========================================================+
+  |     id     |   yes    | The server id to clone delivery service to.            |
+  +------------+----------+--------------------------------------------------------+
+  
+  **Response Properties**
+
+  +------------+----------+--------------------------------------------------------+
+  |  Parameter |   Type   |           Description                                  |
+  +============+==========+========================================================+
+  | hostName   |  string  | The server name to clone delivery service to.          |
+  +------------+----------+--------------------------------------------------------+
+  | dsAssigned |  array   | The deliveryservice name array the server is assigned. |
+  +------------+----------+--------------------------------------------------------+
+
+  **Response Example** ::
+
+    {
+      "response": {
+          "hostName": "atlanta-edge-01",
+          "dsAssigned": [ "test-ds1", "test-ds2" ]
+      }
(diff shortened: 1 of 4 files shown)


traffic_ops/app/lib/API/DeliveryServiceServer.pm
```diff
@@ -67,4 +67,63 @@ sub domains {
...
+
+sub clone_siblings_ds {
+	my $self = shift;
+	if ( !&is_oper($self) ) {
+		return $self->forbidden();
+	}
+
+	my $host_id = $self->param('id');
+	my $server = $self->db->resultset('Server')->find( { id => $host_id } );
+	if ( !defined($server) || $server->type->name ne 'EDGE' ) {
```
mitchell852: looking at https://github.com/Comcast/traffic_control/wiki/Traffic-Ops-API-Guidelines-v1.x

I think maybe this makes more sense:

if ( !defined($server) ) {
 	return $self->not_found(); // 404 is good here
} else if ($server->type->name ne 'EDGE') {
	// return a 403 with a message like 'only can assign delivery services to edge caches'
}

docs/source/development/traffic_ops_api/v12/server.rst
```diff
@@ -780,3 +780,40 @@ Server
   +------------------+--------+------------------------------------------------+
   
 |
+
```
mitchell852: it wasn't intuitive to me what this route was doing until i read through the code. i get what you're trying to do - assign the same delivery services to all caches in a cachegroup....based on the delivery service assignments of one cache. what do you think of this route instead?

POST /api/version/cachegroups/{:id}/assign_delivery_services { "deliveryServices": [ "foo-ds", "bar-ds" ] }

of course this would be a little bit more work as you'd have to do this:

GET /api/1.2/servers/hostname/:name/details.json <-- this will give you the cachegroup id for a server and the delivery services assigned to the server and then you could call

POST /api/cachegroups/id/assign_delivery_services { "deliveryServices": [ "foo-ds", "bar-ds" ] }

in my opinion, that would make the route more intuitive and could be used to not only clone the ds's of a server but also to do simple ds assignments across a cache group. something to think about...

jifyang: Explanation of this route. It involves 3 main items:

server_to: the server to clone delivery service to; its id is the :id in the API.
server_from: the servers where clone the delivery services from.
delivery_service: the related delivery services.

The relationship between them:

server_to <==> server_from <==> delivery_service

server_from are the servers in the same cache group with server_to.
delivery_service are the delivery service that are already assigned to any one of server_from.
The result is: assign delivery_service to server_to.

traffic_ops/app/lib/API/DeliveryServiceServer.pm
jifyang: Accept. Thank you!

jifyang: Updated the code in the pull request.

mitchell852:
Closing as this functionality will be provided via this api route - POST /api/1.2/cachegroups/:id/deliveryservices

Issue #1201, opened 2016-03-22
TO GO-Client Refactor, adds for Ansible


jheitz200:
- Look into renaming from Client to Go-Client
- Code review
- Make sure Go-Client meets all ansible needs. (Get Users, Get Servers, etc)
- Readme

PR #1329, opened 2016-04-20
Client refactor

jheitz200:
Running go vet, golint, go fmt
Adding tests
removing duplicate code
adding code to retrieve users and servers by type from traffic ops

Fixes: #1201


dneuman64:
Congratulations, you are now the "owner" of the traffic ops client code 👍

PR #1343, opened 2016-04-25
Add TO2.0 API Url Rewriter Dockerfile

robert-butts:
This Dockerfile creates a Docker image with NGINX configured to
rewrite requests of a format we want, to that expected by PostgREST.


traffic_ops/experimental/url-rewriter-nginx/Url_Rewriter_Dockerfile
```diff
@@ -0,0 +1,40 @@
+############################################################
+# Dockerfile to build Traffic Ops 1.4 container images
```
mtorluemke: Perhaps this needs updating?

robert-butts: Fixed in 923d5ee5f6ab74e438fec7f47deb2b5d8bcc3c46

Issue #1336, opened 2016-04-21
https (or http + https) deliveryservices with fully qualified HOST_REGEXP entries don't get a remap line


knutsel:
See https://github.com/Comcast/traffic_control/blob/master/traffic_ops/app/lib/UI/ConfigFiles.pm#L255 - that can never work.

PR #1350, opened 2016-04-26
Fix remap.config generation for protocols 1 and 2.

dg4prez:


knutsel:
this fixes #1336 - thanks Derek!

PR #1361, opened 2016-04-28
Update TO default profiles for 1.5

robert-butts:
I'll cherry-pick to `1.5.x` as soon as it's merged.


robert-butts:
Actually, this added comcast.net URLs for `https://odol-trafficstats-01.sys.comcast.net`. Looks like those need added to the get_parameters blacklist. 

Should we remove the parameters entirely? Or leave them in the default profiles, as empty strings or `example.net` or something?

PR #1367, opened 2016-04-29
Fixes issues caused by coexistence of relative/absolute/URI based paths.

elsloo:


traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/FederationsWatcher.java
```diff
@@ -93,14 +93,16 @@ public void configure(final JSONObject config) {
...
+        final File existingDB = databasesDirectory.resolve(databaseName).toFile();
 
         if (!existingDB.exists() || !existingDB.canRead()) {
             return false;
         }
 
         final char[] jsonData = new char[(int) existingDB.length()];
-        new FileReader(existingDB).read(jsonData);
+        final FileReader reader = new FileReader(existingDB);
+        reader.read(jsonData);
```
trevorackerman: It's possible that reader.read will throw an IOException so we need to close it in a try ... finally clause

`try {
    reader.read(jsonData);
} finally {
   reader.close()
}`

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/monitor/TrafficMonitorWatcher.java
```diff
@@ -277,10 +274,18 @@ public void processConfig() {
 		try {
 			String hostList = System.getenv("TRAFFIC_MONITOR_HOSTS");
 
-			final File trafficMonitorConfigFile = new File(propertiesDirectory, monitorProperties);
+			File trafficMonitorConfigFile;
+
+			if (monitorProperties.matches("^\\w+:.*")) {
```
trevorackerman: How about ...
`
File trafficMonitorConfigFile = null

try {
  trafficMonitorConfigFile = new File(new URI(monitorProperties));
} catch (URISyntaxException e) {
  trafficMonitorConfigFile = new File(monitorProperties)
}
`

I agree having a `:` character in a filename seems out of place but it may be better to let the URI class try and validate monitorProperties first

elsloo: I specifically tried that, but the exception was not handled/caught where we expected.

Issue #1344, opened 2016-04-25
Traffic Ops doesn't allow for active/standby configuration in Multi Site Origin.


knutsel:
Thought there was an issue on this already, but can't find it... This is needed for the redundant  linear video origin setup

See code starting @ https://github.com/Comcast/traffic_control/blob/master/traffic_ops/app/lib/UI/ConfigFiles.pm#L1088 and line 1100 

We are hardcoded to `round_robin=consistent_hash`

for active/standby, we need `round_robin=false`

@limited - is this something you guys are going to do??? (referring to our email convo)... Any idea on when?

limited:
In progress as we speak.

knutsel:
Don't know why we can't assign weinfensh, will ask around.

mtorluemke:
Added weifensh as a "collaborator" to the project, now we can assign.

jrushford:
Jan,

There are actually four possible values for ‘round_robin’.  Should traffic ops not support all four?  Here are the possible values for ‘round_robin’:


  *   ‘false’ - The first parent is always chosen if marked available.  Other parents in the list are chosen if the first selected parent is unavailable or if there is a connection error, then another parent is chosen to retry the transaction.
  *   ‘true’ - Parents are chosen using on modulus arithmetic of the client IP.  As above, if the first chosen parent is unavailable or if there is a connection error, another parent is chosen to retry the transaction.
  *   ‘strict’ - Parents are chosen in strict round robin fashion.  As above, if the first chosen parent is unavailable or if there is a connection error, another parent is chosen to retry the transaction.
  *   ‘consistent_hash’ - Parents are chosen from a consistent hash ring based upon a hash of the request url.  As above, if the first chosen parent is unavailable or if there is a connection error, another parent is chosen to retry the transaction.

Thanks
--
John J. Rushford
IPCDN Engineering
1400 Wewatta Street, Denver Colorado 80202
John_Rushford@cable.comcast.com

jianli123:
The design of Multi Site Origin Design For Algorithm Primary/backup is ready at https://docs.google.com/document/d/1TAo2jBZqdb8TY9Tn_Kjn1zYm7XHt5jhQnt1AvBK8eO0/edit?usp=sharing and please take a look.  And the code will come soon. thanks

smalenfant:
Is there anyway we can get away from the single profile per origin? Could we separate the "cache server assignment" from the "origin server assignments"?

[...]

PR #1377, opened 2016-05-03
add mso algorithm support on traffic ops

weifensh:


weifensh:
#1344

weifensh:
will update file name of 20160430000000_add_multisite_algrithom.sql when this PR is reviewed

weifensh:
@mitchell852 @mtorluemke Could you help to review? Thanks:)

mitchell852:
yep, i'm reviewing today

mitchell852:
just curious. why not just store the text values (consistent_hash, false, strict, true) in the database for multi_site_origin_algorithm rather than using the numbers 0,1,2,3. when querying the database sometimes it's hard to remember what 2 stands for...for example...

docs/source/admin/quick_howto/multi_site.rst
```diff
@@ -38,27 +38,55 @@ Configure Multi Site Origin
 	:scale: 100%
 	:align: center
 
-4) assign the org servers to the delivery service that will have the multi site feature
+4) Assign the org servers to the delivery service that will have the multi site feature:
```
mitchell852: I would switch step#4 and step#5 because if you're creating a new delivery service,  you can't do step 4 until you do step 5

weifensh: @mitchell852 yeah, I am switching them, thanks.

weifensh:
@mitchell852 Thanks for your comments. Yes, it looks to be more direct to store the text (consistent_hash, false, strict, true) in the database. Initially I was following the "Protocol" field with values of "http", "https", and etc., which is stored as 0, 1, 2... Now I get that mso algorithm is a bit different story so your suggestion looks better. But the change from number to text is big. If the current implementation is ok, could we use it? :-) Of course if you prefer to use text value, I will modify it in one or two days. Thanks.

mitchell852:
i think it's fine to use 0..4, plus it's consistent with other implementations (i.e. protocol as you mentioned). thx.

traffic_ops/app/db/migrations/20160628000000_add_multisite_algrithom.sql
```diff
@@ -0,0 +1,24 @@
+/*
```
mitchell852: can you rename this file? looks like there's already a 20160628000000....sql

(not your fault. happens)

mitchell852:
can you run perltidy against a bunch of your files. looks like your formatting is a bit off.

weifensh:
@mitchell852 Just renamed the goose sql file name. And also I reformatted the perl files (except Result/Deliveryservice.pm) with perltidy. Please help to review if they are good. Thanks.

limited:
Does this fix #1377 ?

mitchell852:
It does but I want to do a little more testing on #1377 and then i'll close it.

dg4prez:
Tested in the lab with all four MSO configuration possibilities, as well as ranking configuration for primary/secondary configuration.  Parent config files formatted as expected.

dg4prez:
I just found a small bug - it looks like we get an entry in the parent.config on the mid for each regex entered into the delivery service.  They're all duplicate entries.

weifensh:
@dg4prez Thanks Derek. This duplicated issue seems to exist before my pull request, but let me verify this later.

PR #1398, opened 2016-05-09
Traffic Router now provides the ability to delegate traffic to subordinate delivery services

trevorackerman:
Also known as steering


dneuman64:
While running this code in the lab I saw a warning log message like the one below about every minute in the log.  We should figure out why this is happening before merging.  I verified the config file was in `/opt/traffic_router/conf`.

> WARN  2016-05-12T02:39:10.818 [pool-8-thread-1] com.comcast.cdn.traffic_control.traffic_router.core.monitor.TrafficMonitorWatcher - Cannot load traffic monitor properties file /$[traffic_monitor.properties] file not found!

elsloo:
I'm not 100% sure, but I think that warning message is because the branch for this change hasn't been rebased to include the properties file changes I made recently.

dneuman64:
That was my thinking too.  I put the comment on here so we would remember to check it out before merging.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -74,12 +71,12 @@
 
 public class TrafficRouter {
 	public static final Logger LOGGER = Logger.getLogger(TrafficRouter.class);
+	public static final String XTC_STEERING_OPTION = "xtc-steering-option";
```
elsloo: Other instances of this are mixed case, so I presume that the use of this string is case insensitive?

Also, we need to change the value of the header name we expect. `XTC-Steering-Option` it should be `X-TC-Steering-Option` or just `TC-Steering-Option`. Originally, the best practice was to prefix a custom header with `X-` but that has since been deprecated via RFC 6648.

PR #1407, opened 2016-05-11
Fixed an issue that prevented updates from being queued on child when…

### LINKED ISSUE #1406: OFFLINE/ONLINE buttons do not cause updates to be queued properly
The shortcut OFFLINE/ONLINE buttons are not causing updates to be queued properly for child caches; the existing code attempts to queue updates for "edges," and just doesn't work.

elsloo:
… a parent's state is administratively changed. This resolves #1406.


mtorluemke:
Awesome!

PR #1416, opened 2016-05-11
create tmp file in same dir as conf

### LINKED ISSUE #1386: TO: traffic_ops postinstall issue
The following code is copied from "traffic_ops/install/bin/postinstall":

        # write whole config to temp file
        my $tmpfile = File::Temp->new();
        print $tmpfile $dumper->Dump();
        close $tmpfile;

        # rename current config file to something unique so it's not lost
        my $backup_num = 0;
        my $backup_name;
        do {
                $backup_num++;
                $backup_name = "$cdn_conf.backup$backup_num";
        } while ( -e $backup_name );
        rename( $cdn_conf, $backup_name ) or die("rename(): $!");

The rename() will fail if the directory "/tmp/" and the directory "traffic_ops/install/bin/" are not in the same disk partition.

The same issue will be found in the script "traffic_ops/install/bin/generateCert"

dangogh:
Fixes #1386 

this guarantees not in separate file system, so no cross-device link error.

To test: `export TMPDIR=<directory in another file system>` and run previous version and then this one.   May need to touch /opt/traffic_ops/.reconfigure to force conf file to be re-created.

### DIFF
--- a/traffic_ops/install/bin/generateCert
+++ b/traffic_ops/install/bin/generateCert
@@ -71,8 +71,8 @@ sub writeCdn_conf {
 	my $dumper = Data::Dumper->new( [$cdnh] );
 	$dumper->Indent(1)->Terse(1)->Quotekeys(0);
 
-	# write whole config to temp file
-	my $tmpfile = File::Temp->new();
+	# write whole config to temp file in pwd (keeps in same filesystem)
+	my $tmpfile = File::Temp->new(DIR => '.');
 	print $tmpfile $dumper->Dump();
 	close $tmpfile;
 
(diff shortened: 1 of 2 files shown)


traffic_ops/install/bin/postinstall
```diff
@@ -390,8 +390,8 @@ sub writeSecret {
 	my $dumper = Data::Dumper->new( [$cdnh] );
 	$dumper->Indent(1)->Terse(1)->Quotekeys(0);
 
-	# write whole config to temp file
-	my $tmpfile = File::Temp->new();
+	# write whole config to temp file in pwd (keeps in same filesystem)
+	my $tmpfile = File::Temp->new(DIR => '.');
```
yorma: In generatedCert in line 74, need to make the same change.

dangogh: thanks..  can you put in a separate issue for that?

traffic_ops/install/bin/generateCert
```diff
@@ -71,8 +71,8 @@ sub writeCdn_conf {
 	my $dumper = Data::Dumper->new( [$cdnh] );
```
dewrich: Probably should remove this Dumper output before we send it through

PR #1444, opened 2016-05-23
TO: Delete config file parameters when deleting Delivery Service

xiaodzha:
Fix bug #1431 

When Delivery Service is deleted,  delete all the config parameters which belong to it.


mtorluemke:
Thanks!

PR #1452, opened 2016-05-24
fixed an issue that caused not all filters to show up for a steering …

dneuman64:
…assignment.


dneuman64:
This was fixed in #1456

PR #1463, opened 2016-05-26
add deploy user role

dg4prez:


traffic_ops/app/db/migrations/20160526140027_add_deploy_role.sql
```diff
@@ -0,0 +1,8 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+insert into role (name, description, priv_level) select * from (select 'deploy', 'Deployment role', 15) as tmp where not exists (select name from role where name = 'deploy') limit 1;
+
+-- +goose Down
+-- SQL section 'Down' is executed when this migration is rolled back
+
```
elsloo: For the goose down, should we add: `delete from role where name = 'deploy' and description = 'Deployment role' and priv_level = 15;` ?

dg4prez: Agreed.  Added and tested.

PR #1464, opened 2016-05-26
Remove TM config.pl unused domainname call

robert-butts:
Fixes Docker, which doesn't set the `domainname`.

We could change Docker to not use `config.pl`, but, since the script was never actually using `domain_name`, removing the variable was the easiest fix.


elsloo:
@smalenfant I don't believe that this change will break your use of this script, but please let us know if it does.

PR #1469, opened 2016-05-27
Adding optional param 'useInTable'

### LINKED ISSUE #1468: TO Client: add optional "UseInTable" param for Types
Need to ability to request TO data types by the table they are used in.

jheitz200:
Adding optional param to get data types based on the table they are used in.

fixes #1468

### DIFF
--- a/traffic_ops/client/type.go
+++ b/traffic_ops/client/type.go
@@ -16,7 +16,10 @@
 
 package client
 
-import "encoding/json"
+import (
+	"encoding/json"
+	"errors"
+)
 
 // TypeResponse ...
 type TypeResponse struct {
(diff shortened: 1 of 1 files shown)


traffic_ops/client/type.go
```diff
@@ -28,10 +31,17 @@ type TypeResponse struct {
 type Type struct {
 	Name        string `json:"name,omitempty"`
 	Description string `json:"description,omitempty"`
+	UseInTable  string `json:"useInTable,omitempt"`
 }
 
 // Types gets an array of Types.
-func (to *Session) Types() ([]Type, error) {
+// optional parameter: userInTable
+func (to *Session) Types(useInTable ...string) ([]Type, error) {
```
robert-butts: That's interesting. It hadn't occurred to me you could use a variadic arg to simulate function overloads.

But, we should at least be aware it's not idiomatic Go: https://golang.org/doc/faq#overloading

jheitz200: Do you have any other ideas to achieve the same result?

robert-butts: Well, you'd have to either make 2 functions with 2 different names, or pass a slice or the empty string, so you'd have to call `Types(nil)` or `Type("")`.

I wasn't saying we shouldn't do it. I'm not convinced we should do everything Rob Pike says "no ifs ands or buts". Just saying we should be aware.

PR #1473, opened 2016-05-30
Fix #1472: Add config parameters when adding delivery service/ cloning cache server

### LINKED ISSUE #1472: TO: Config Parameters won't be created when adding a delivery service
1. Create a delivery service with 'Regex remap expression'  & 'Cache URL expression'  filled and assign an edge cachegroup to this delivery service

2. Two parameters should be created whose config_file should be  cacheurl_<ds_xml_id>.config & regex_remap_<ds_xml_id>.config

3. There are no parameters created.

xiaodzha:
Fix #1472


mtorluemke:
Thanks!

PR #1485, opened 2016-06-02
To steering api updates

dneuman64:
Added ability to update and remove filters.
Traffic Ops now returns a 400 (instead of a 200) when you try to update a steering assignment with invalid json.

### DIFF
--- a/traffic_ops/app/lib/API/DeliveryService/Steering.pm
+++ b/traffic_ops/app/lib/API/DeliveryService/Steering.pm
@@ -201,7 +201,7 @@ sub update() {
     }
 
     if (!$self->req->json->{'targets'} ) {
-        return $self->render(json => {"message" => "please provide a valid json"}, status => 400);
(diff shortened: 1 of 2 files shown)


trevorackerman:
I noticed that one of the tests still accepts the wrong json PUT data

https://github.com/dneuman64/traffic_control/blob/be93fd85ac0e4f8cef322c446523a419875d5dc4/traffic_ops/app/t/api/1.2/steering_internal.t#L276

The 'filters' attribute is in the wrong place. Can you remove just that one line and have the test assert that we get back a 400? i.e. filters is a required attribute for the PUT data.

The test you came up with for removing filters looks good!
https://github.com/Comcast/traffic_control/pull/1485/files#diff-27ed8be6da922277af8c7baa33b71575R337

dneuman64:
hmm, that is not anything I changed, but I will remove the filters from there.  I should probably have a check to make sure weight and deliveryService is included for target.
As for filters, they are not required.  it is perfectly acceptable to submit a json without filters, and I already have tests to validate accordingly.

PR #1500, opened 2016-06-07
Centos7 ort integration.

jrushford:
This pull request contains changes to traffic_ops_ort.pl following integration testing on a CentOS 7 machine.  Changes are principally around the system V init script commands service and chkconfig and the systemd commands used in centos 7.  These changes were made to fix centos 7 integration issues.  I think that the traffic ops chkconfig endpoint needs to be generalized to support both chkconfig and systemd.  Following this, these changes should be refactored to better abstract systemd and chkconfig differences.

NOTE:  Centos 6 regression testing is in progress.  I'll update this PR with regression test results.

### DIFF
--- a/traffic_ops/bin/traffic_ops_ort.pl
+++ b/traffic_ops/bin/traffic_ops_ort.pl
@@ -100,6 +100,9 @@ my $ERROR = 2;
 my $FATAL = 1;
 my $NONE  = 0;
 
+my $RELEASE = &os_version();
+( $log_level >> $DEBUG ) && print "DEBUG OS release is $RELEASE.\n";
+
 my $script_mode = &check_script_mode();
 &check_run_user();
 &check_only_copy_running();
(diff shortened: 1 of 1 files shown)


jrushford:
Ran Centos6 regression tests on odol-qa-atsec-01.cdnlab.comcast.net and found no issues.

traffic_ops/bin/traffic_ops_ort.pl
```diff
@@ -326,9 +336,73 @@ sub process_cfg_file {
...
+}
+
+sub systemd_service_status {
+	my $pkg_name = shift;
+	my $running_string;
+	my $running = 0;
+	my $pid;
+	my $prog;
+
+	open(FH, "/sbin/service $pkg_name status|") or die ("/bin/service: $!");
```
limited: To be consistent with other uses of systemctl, should this be '/usr/bin/systemctl status $pkgname'?

jrushford: 'limited', I made the change.

limited: Great thanks Jon!
--Eric

PR #1522, opened 2016-06-13
TO: Feature/add server guid

### LINKED ISSUE #1520: TO: Add GUID to server table
add GUID text column to hold id of server within a cloud provider, e.g. openstack.  This id should be available from GET/POST/PUT apis and from gui.

dangogh:
feature requested by @dewrich -- Fixes #1520

### DIFF
--- a/traffic_ops/app/db/migrations/20160613153313_add_guid.sql
+++ b/traffic_ops/app/db/migrations/20160613153313_add_guid.sql
@@ -1,0 +1,8 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+alter table server add column guid varchar(45) DEFAULT NULL AFTER router_port_name;
+
+-- +goose Down
+-- SQL section 'Down' is executed when this migration is rolled back
+alter table server drop column guid;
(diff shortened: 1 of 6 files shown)


traffic_ops/app/db/migrations/20160613153313_add_guid.sql
```diff
@@ -0,0 +1,8 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+alter table server add column guid varchar(45) DEFAULT NULL AFTER router_port_name;
```
mtorluemke: Should we name as UUID? Does it matter?

dangogh: I don't think it does matter -- guid is what @dewrich requested, and I didn't see any compelling reason to make it different.   Easily changed, though, if done soon..

smalenfant:
How could we extend the server table to have key/value tags instead? The problem today is that if different organization needs a new field, you have to add it in the UI, extend the database columns, etc... Adding simple tags could possibly fix all this at once. This sounds more like an enhancement.

Issue #1449, opened 2016-05-24
Traffic Router fetches invalid czfile/rgbfile to its db directory


weifensh:
If the parameter "coveragezone.polling.url" of TR points to an invalid coveragezone file, TR still downloads it to its local db directory. When TR restarts, there will be no valid coveragezone file to use.

A possible fix is not to download invalid files to local db. Thus TR could use the latest valid config file.

dneuman64:
This was fixed by #1748

PR #1541, opened 2016-06-22
Fix TR download invalid db issue

weifensh:
Create a backup when downloading a config file from HTTP server;
If the downloaded one is invalid, revert to the backup file;
Try to delete all temp files;

PS: I was unable to resolve the PMD.CyclomaticComplexity in the new method loadNewDatabase(), so it was suppressed.


weifensh:
Linking to issue #1449

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
```diff
@@ -151,28 +149,102 @@ public boolean updateDatabase() {
...
+	public boolean loadDatabaseWrap() {
+		try {
+			if (loadDatabase()) {
+				setLoaded(true);
+				databaseReverted = false;
+				trafficRouterManager.trackEvent("last" + getClass().getSimpleName() + "Update");
+				return true;
+			}
+		} catch (Exception e) {
+			LOGGER.error("loadDatabase with exception", e);
```
dneuman64: Looks like the old logging message was `LOGGER.error("Failed to load existing database! " + e.getMessage());`, we should keep it consistent.

I think we want to return false here

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
```diff
@@ -151,28 +149,102 @@ public boolean updateDatabase() {
...
+		}
+
+		LOGGER.warn("loadDatabase failed " + getClass().getSimpleName());
```
dneuman64: do we want this to be logged as an error?

trevorackerman:
Please add an automated unit test that will be run by maven that demonstrates the error.
Based on the issue you submitted the test should confirm that 

- the remote resources is saved on the file system somewhere other than the local db
- the temporary file the resource is saved to is moved to the local db only when it is deemed valid

Making a backup of the existing database file and then having all the logic to juggle the backup depending on whether things were successful or not is complicated. It'd be better to simply save the new data to a temporary location and only use the data and overwrite the existing file when the verify method returns true.

Thanks for working on this, we do really appreciate your contributions 😃

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
trevorackerman: This will fall through to the end of the routine so false will be returned

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
trevorackerman: It's better without this log line. The only way we can get here is if an exception is thrown above and that already produces a log entry. So let's remove this one.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
```diff
@@ -272,29 +344,21 @@ private boolean compareFiles(final File a, final File b) throws IOException {
 		return false;
 	}
 
-	protected boolean copyDatabaseIfDifferent(final File existingDB, final File newDB) throws IOException {
+	static class CopyDatabaseException extends Exception { }
```
trevorackerman: Let's just use booleans instead of creating exception classes and throwing them.
Also let's not jam static classes inside existing ones in newly introduced code.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
```diff
@@ -272,29 +344,21 @@ private boolean compareFiles(final File a, final File b) throws IOException {
...
-			if (existingDB.isDirectory()) {
-				for (File file : existingDB.listFiles()) {
-					file.delete();
-				}
-				LOGGER.debug("[" + getClass().getSimpleName() + "] Successfully deleted database under: " + existingDB);
-			} else {
-				existingDB.delete();
+			if (!existingDB.renameTo(backupDB)) {
+				LOGGER.fatal("Unable to rename " + existingDB + " to backup " + backupDB + "; current working directory is " + System.getProperty("user.dir"));
+				throw new CopyDatabaseException();
```
trevorackerman: Just return false here instead of throwing an exception

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
```diff
@@ -304,26 +368,14 @@ protected boolean copyDatabaseIfDifferent(final File existingDB, final File newD
 
 		if (!renamed) {
 			LOGGER.fatal("[" + getClass().getSimpleName() + "] Unable to rename " + newDB + " to " + existingDB.getAbsolutePath() + "; current working directory is " + System.getProperty("user.dir"));
-			return false;
+			backupDB.renameTo(existingDB); // revert backup to existing
+			throw new CopyDatabaseException();
```
trevorackerman: Just return false here and leave out the comment.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/NetworkNode.java
```diff
@@ -95,6 +95,7 @@ public static NetworkNode generateTree(final JSONObject json) {
                             root.add6(new NetworkNode(ip, loc, geolocation));
                         } catch (NetworkNodeException ex) {
                             LOGGER.error(ex, ex);
+                            return null;
```
trevorackerman: Can you express in a test what will happen that is incorrect if we do not return null here?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/RegionalGeoUpdater.java
```diff
@@ -33,7 +33,13 @@ public RegionalGeoUpdater() {
 
     public boolean loadDatabase() throws IOException {
         final File existingDB = databasesDirectory.resolve(databaseName).toFile();
-        RegionalGeo.parseConfigFile(existingDB);
+        if (!existingDB.exists() || !existingDB.canRead()) {
+            return false;
+        }
+
+        if (!RegionalGeo.parseConfigFile(existingDB)) {
```
trevorackerman: Simplify to 
`return RegionalGeo.parseConfigFile(existingDB);`

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/NetworkUpdater.java
```diff
@@ -36,7 +36,9 @@ public boolean loadDatabase() throws IOException, JSONException {
 			return false;
 		}
 
-		NetworkNode.generateTree(existingDB);
+		if (null == NetworkNode.generateTree(existingDB)) {
```
trevorackerman: simplify to
`return NetworkNode.generateTree(existingDB)) != null`

dneuman64:
@weifensh, we are planning on cutting a 1.6 release on 6/30, please let us know if you think you can take a look at Trevor's suggestions by then or if we should move this PR to the 2.0 release (end of August).

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/NetworkUpdater.java
weifensh: sure

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/RegionalGeoUpdater.java
weifensh: sure, thanks

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
weifensh: @dneuman64 In my test, I haven't configured SteeringWatcher yet, but it still reports this error. If configuring SteeringWatcher is optional, I just put this message a warn one.
@trevorackerman Yes, in most cases loadDatabaseWrap() return false because of the Exception, but in some cases loadDatabase() returns false, I tried to log that case if it does happen.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/NetworkNode.java
weifensh: Adding itRejectsInvalidIpV4Network() & itRejectsInvalidIpV6Network().

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
dneuman64: @weifensh Ok, lets keep it as a warning then.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
weifensh: Yeah, initially I did want to avoid the Exception of CopyDatabaseException. But I wanted to return another status that the copy operation is failed, to distinguish with the the status of the same file or not. Maybe returning an enum to indicate three status is another option (e.g. Same/Different/Failed). Any other suggestion? Thanks.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
weifensh: @ trevorackerman Thanks for your valuable suggestions. Not knowing how to reply your general comments below, so I am replying here:) 

Regarding the UT, I will do some research and hope I could add it tomorrow.

And yes I did try to save the downloaded file to a temp place, verify it, and then decide to move to local db or not. But the change appears to be big, so I chose the copy/revert operation, which does look a bit tricky. Let me try to go through this tomorrow again, to see if we can copy a db only when it is valid.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
trevorackerman: Then I would recommend splitting copyDatabaseIfDifferent method in two methods that both return boolean and no need to return an exception.

- isDatabaseDifferent
- swapDatabase

I'm suggesting swapDatabase instead of copyDatabase because traffic router is doing move filesystem operations instead of copy filesystem operations.

As usual log anything unusual that you believe strongly warrants an operator's attention.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
weifensh: @trevorackerman Take a look at the code again and it appears that copy a db only when it is valid shall be feasible. It will be like:
1. Download the db to a temp place if it is a new one;
2. Try to load it from the temp place;
3. If valid and OK, replace the current db with the downloaded temp one;
4. Otherwise, just delete the temp file.

Potential changes would be:
change loadDatabase() to loadDatabase(final File newDB);
add new method GeolocationService::reloadDatabase(final File newDB) to load db from alternative places...

How do you think? If you agree, I might start to revise the PR soon after June 30th.

trevorackerman: Sounds like a good plan! Thanks for your efforts to make traffic router better.

dneuman64:
@weifensh Sorry we havent looked at this for a while, can you resolve the conflicts so we can get this into the 1.7 release?  Thanks!

weifensh:
@dneuman64 I am revising this PR, hope to finish it in one or two days. I have a question on the neustar DB, is there any guide on how to use it? Because this PR will the touch the code related to neustar, but I have not found any doc on how to configure it and don't know how to test it. Could you help to point me how to configure? many thanks.

dneuman64:
What in the neustar package are you needing to change?  I don't think you can just configure it without having a copy of the database which you need to get from neustar.  There is a [README.md ](https://github.com/Comcast/traffic_control/blob/master/traffic_router/neustar/README.md) in the neustar package that has some more information on getting it setup.  If you are just needing to make changes to geoLocationService then hopifully our unit tests will cover you, if not I can do some manual testing with neustar and make sure things are still happy.

weifensh:
@dneuman64 Thanks David. It appears that NeustarDatabaseUpdater has its own logic to download db, so my fix here will not touch its code. I will submit my revised PR tomorrow.

weifensh:
Closing this PR, to open another one #1748

Issue #1509, opened 2016-06-08
TO: POST/PUT json body keys should be camelCase, not snake_case


dangogh:
the json body for all POST/PUT requests should have keys in camelCase and not snake_case.   Most of these were added in #1142.   This applies to POST and PUT on /servers, /deliveryservices, /profiles. /divisions, /regions, and /cachegroups (I think that's all of them...)

Example:  `POST .../api/1.2/servers` will accept this json (from `traffic_ops/t/api/1.2/server_create.t`):
>  {
>          "host_name" => "tc1_ats2",
>          "domain_name" => "northbound.com",
>          "cachegroup" => "mid-northeast-group",
>          "cdn_name" => "cdn1",
>          "interface_name" => "eth0",
>          "ip_address" => "10.74.27.184",
>          "ip_netmask" => "255.255.255.0",
>          "ip_gateway" => "10.74.27.1",
>          "interface_mtu" => "1500",
>          "phys_location" => "HotAtlanta",
>          "type" => "MID",
>          "profile" => "MID1"
>  }

but it should look like this:
>  {
>          "hostName" => "tc1_ats2",
>          "domainName" => "northbound.com",
>          "cachegroup" => "mid-northeast-group",
>          "cdnName" => "cdn1",
>          "interfaceName" => "eth0",
>          "ipAddress" => "10.74.27.184",
>          "ipNetmask" => "255.255.255.0",
>          "ipGateway" => "10.74.27.1",
>          "interfaceMtu" => "1500",
>          "physLocation" => "HotAtlanta",
>          "type" => "MID",
>          "profile" => "MID1"
>  }

limited:
I think we already have a patch for this. Perhaps it just hasn't made its way into open source yet.

dangogh:
if that can make it's way in here,  I'd really appreciate it..   thanks!

yorma:
My colleagues and I will work on this by different api, thanks.

dangogh:
Can you describe how you're planning to do that?  I just want to make sure we know what's coming..

yorma:
We plan to check APIs we pulled if the keys are snake_case, we will change to camelCase. It will be divided into tasks, so only 1 or 2 APIs in one pull request. We may start next week or the week after next week base on the timing. Thanks,

mitchell852:
keep in mind that release 1.6 will be cut on 6/30.

dangogh:
[...]

PR #1550, opened 2016-06-27
change POST/PUT server API json body keys to camelCase

zhilhuan:
fix for #1509

### DIFF
--- a/traffic_ops/app/lib/API/Server.pm
+++ b/traffic_ops/app/lib/API/Server.pm
@@ -401,73 +401,89 @@ sub check_server_params {
 	my $update_base = shift;
 	my %params      = %{$json};
 	my $err         = undef;
+	my %errFields   = ();
 
-	if ( defined( $json->{'interface_mtu'} ) ) {
-		if ( $json->{'interface_mtu'} != '1500' && $json->{'interface_mtu'} != '9000' ) {
-			return ( \%params, "'interface_mtu' '$json->{'interface_mtu'}' not equal to 1500 or 9000!" );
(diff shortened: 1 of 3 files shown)


traffic_ops/app/lib/API/Server.pm
```diff
@@ -898,14 +914,15 @@ sub update {
 	# this just creates the log string for the log table / tab.
 	my $lstring = "Update server " . $update->host_name . " ";
 	foreach my $col ( keys %{ $org_server->{_column_data} } ) {
-		if ( defined( $params->{$col} )
-			&& $params->{$col} ne ( $org_server->{_column_data}->{$col} // "" ) )
+        my $colParam = $col =~ s/_(\w)/\U$1/gr;
```
dangogh: This line gives me a syntax error (from `perl -c lib/API/Server.pm`)
```perl
Bareword found where operator expected at lib/API/Server.pm line 917, near "s/_(\w)/\U$1/gr"
syntax error at lib/API/Server.pm line 917, near "s/_(\w)/\U$1/gr"```
I think you want `(my $colParam = $col) =~ ...`   or use `Mojo::Util::camelize`

zhilhuan: This is due to "/r", which is a regex feature introduced in perl 5.14.0. Will get ride of this in the following commit.

dangogh:
Thanks..   I wasn't aware of that feature, and 5.10 is the version shipped with CentOS 6.7,  so it wouldn't have that.. Checking again now.

PR #1554, opened 2016-06-27
crconfig parameter consistency check

### LINKED ISSUE #1521: Server CDN assignment and profile assignment must take the domain_name profile parameter into account
Server profile assignment and CDN assignments must align, or the current code in `UI/Topology.pm` might use an incorrect domain_name value when generating the CRConfig. Specifically if a server is incorrectly assigned to a profile that belongs to a different CDN, and that profile has a domain_name parameter, it's possible for this incorrect FQDN to be selected as the top level domain for the entire CDN. Conversely, if the wrong CDN is selected for a given server, the same thing can happen, as a server assigned to the other CDN's profile will appear in the list of servers when generating the CRConfig.

This is due to the following code in `UI/Topology.pm` (note my comments added for clarity):

```
    # elsloo: this loads all profiles for servers assigned to this CDN..
    # ..if the profile is for another CDN with a domain_name parameter that doesn't belong to this CDN..
    my $rs_cdn_profiles = $self->db->resultset('Server')->search(
        { 'cdn.name' => $cdn_name },
        {
            select => [ 'cdn.id', 'me.profile', 'me.type' ],
            join   => 'cdn',
            prefetch => [ 'profile', 'type' ],
            distinct => 1
        }
    );

    while ( my $row = $rs_cdn_profiles->next ) {
        push( @{ $profile_cache->{ $row->type->name } }, $row->profile->id );
        $cdn_id = defined($cdn_id) ? next : $row->cdn->id;
    }

    my %param_cache;
    my @profile_caches;

    ...

    # elsloo: all profiles and one that may have a bad domain_name parameter are in @profile_caches
    my %condition = (
        -and => [
            profile                 => { -in => \@profile_caches, },
            'parameter.config_file' => 'CRConfig.json'
        ]
    );
    # elsloo: use the above condition to find all profile parameters
    my $rs_pp = $self->db->resultset('ProfileParameter')->search( \%condition, { prefetch => [ { 'parameter' => undef }, { 'profile' => undef } ] } );

    ...

    while ( my $row = $rs_pp->next ) {
        ...
        # elsloo: note $ccr_domain_name and the use in $data_obj - there's no break in this loop..
        # ..which means a race with the domain_name profile parameter, and if the last domain_name is bad..
        if ( $row->parameter->name eq 'domain_name' ) {
            $ccr_domain_name = $row->parameter->value;
            $data_obj->{'config'}->{ $row->parameter->name } = $row->parameter->value;
        }
        ...
    }
```

Note that `$ccr_domain_name` is used to generate FQDNs for all DSs on each cache object and `            $data_obj->{'config'}->{'domain_name'}` is used to configure the CDN TLD on the Traffic Router.

If the server is assigned to a profile or CDN that do not match, the CRConfig could be configured with the wrong CDN TLD and DS FQDNs, which will propagate to the DNS side of Traffic Router, which will cause Traffic Router to stop serving DNS for the desired TLD/domains and serve DNS for the incorrect `domain_name`.

dangogh:
Fixes #1521 -- before generating CRConfig, checks certain params (including domain_name) pulled from profiles on all servers in the CDN.   Generates error if any inconsistency found.

### DIFF
--- a/traffic_ops/app/lib/UI/Tools.pm
+++ b/traffic_ops/app/lib/UI/Tools.pm
@@ -1,4 +1,5 @@
 package UI::Tools;
+
 #
 # Copyright 2015 Comcast Cable Communications Management, LLC
 #
@@ -29,7 +30,6 @@ use HTTP::Cookies;
 sub tools {
     my $self = shift;
 
-    print "hahy\n";
     &navbarpage($self);
     my %serverselect;
     my $rs_server = $self->db->resultset('Server')->search( undef, { columns => [qw/id host_name domain_name/], orderby => "host_name" } );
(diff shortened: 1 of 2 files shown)


dangogh:
this really needs to get into 1.6..   comments welcome..

PR #1555, opened 2016-06-27
Updated server scripted dashboard to include telegraf data

dneuman64:
If you have telegraf data for a server (specifically cpu, mem, system, and diskio) you will now get some new graphs when you click on the server graph.  If you do not have telegraf data for the cache then you will get empty graphs.  

Also fixed an issue where connections were getting multiplied by 1000.

The new graphs that show are CPU usage, Memory Usage, Load Average, and Read/Write time.


dneuman64:
This closes #1555

PR #1557, opened 2016-06-28
Fix allow_ip6 issue in astats plugin

weifensh:
IPv6 address is 128bit/16byte long;
AF_INET to AF_INET6 for inet_pton calls;
some other changes.

### DIFF
--- a/traffic_server/plugins/astats_over_http/astats_over_http.c
+++ b/traffic_server/plugins/astats_over_http/astats_over_http.c
@@ -538,7 +538,7 @@ void TSPluginInit(int argc, const char *argv[]) {
 	info.support_email = "justin@fp-x.com";
 	astatsLoad = time(NULL);
 
-	if (TSPluginRegister(&info) != TS_SUCCESS) {
+	if (TSPluginRegister(TS_SDK_VERSION_3_0, &info) != TS_SUCCESS) {
         TSError("Plugin registration failed. \n");
 	}
 
@@ -582,9 +582,9 @@ static bool is_ip_allowed(const config_t* config, const struct sockaddr* addr) {
 	if(!addr) {
 		return true;
 	}
-	const char *ip = addr->sa_data+2;
 
 	if (addr->sa_family == AF_INET && config->allowIps) {
+	    const char *ip = (const char*)&((struct sockaddr_in*)addr)->sin_addr;
 		for(i=0; i < config->ipCount; i++) {
 			ipmask = config->allowIps+ (i*5);
 			if(is_ip_match(ip, ipmask, ipmask[4])) {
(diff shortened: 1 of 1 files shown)


weifensh:
linking to issue #1557

traffic_server/plugins/astats_over_http/astats_over_http.c
```diff
@@ -538,7 +538,7 @@ void TSPluginInit(int argc, const char *argv[]) {
 	info.support_email = "justin@fp-x.com";
 	astatsLoad = time(NULL);
 
-	if (TSPluginRegister(&info) != TS_SUCCESS) {
+	if (TSPluginRegister(TS_SDK_VERSION_3_0, &info) != TS_SUCCESS) {
```
PSUdaemon: Is this related to ipv6 changes?

PSUdaemon:
I have concerns about this. The repo you are using for this is not the canonical one. I'd also like to obsolete this plugin in favor of the `stats_over_http` plugin from the main ATS repo and a new `remap_stats` like plugin that just adds the few stats this plugin has manually added. I'd be much more interested in seeing that work completed instead of further work on this one. Is this something you might be interested in helping out with?

limited:
Thanks for the review Phil-
  What is the canonical repo for the astats plugin?

PSUdaemon:
It's an internal Comcast one that is branched off of the `trafficserver` repo :). This is why I'd like to change all that. The `astats` plugin is heavily overlapping with `stats_over_http` and likely if we made a `system_stats` plugin much like the `remap_stats` plugin we could get it into the trafficserver mainline. It really doesn't belong in a Traffic-Control repo at all, imo.

limited:
I agree `astats` is out of Traffic Control's scope and would be interested in seeing TC use `stats_over_http` instead. 

If we are re-examining module boundaries, does it still make sense for Traffic Server to report on system stats?  If we ever decide to monitor something other than bandwidth or load average, we'd need to come out with a new build of `ats`. Have you thought at all about using some other monitoring tool  to poll the system health instead of binding it to `astats`? 

In thinking about this, it could be useful in a few different cases:
1. If a cache other than `trafficserver` is used, we don't need to reimplement system stats gathering.
2. We haven't observed our Traffic Routers getting overloaded, but this could open the door for health monitoring and load balancing amongst TRs as well.

The downside is of course more packages to install, more configuration to push down and one more thing to break.

PSUdaemon:
I think what we are both proposing is compatible. If we end up with a separate "system stats" plugin that added that stuff, then it can be swapped with any other sort of thing given appropriate support in traffic monitor. This would require more work, but it's feasible. I think what I am proposing is something that can be accomplished easier, as it's a slight renaming of variables. Then we can make it more generic down the road to meet your proposal's requirements. It would be more flexible and people could go as far as they wanted. For example, integrate with their already existing monitoring which means fewer packages instead of more.

limited:
Yeah, I think this is probably step in the right direction. I like that its a more incremental change with some possibilities for future improvement. 

To use a new `system_stats` plugin, does TM make a separate HTTP call to the cache to get the system stats? Or will the system stats be included as part of the `stats` call?

dneuman64:
I think we should get a new issue opened up for getting rid of astats in favor of something else...

PSUdaemon:
It will register new plugin stats much like the `remap_stats` plugin does.

https://github.com/apache/trafficserver/blob/master/plugins/experimental/remap_stats/remap_stats.c#L45-L90

And those stats will end up in the output of `stats_over_http`. So there would be some "schedule every" type even handler that looks at the system interfaces like `astats` does and then update the internal stats that are then exposed via `stats_over_http`.

traffic_server/plugins/astats_over_http/astats_over_http.c
weifensh: No, it's not related to ipv6. This change just helps astats to get compiled with ATS 5.3.2.

dneuman64:
can this be closed since we are going to go a different direction?

PR #1565, opened 2016-06-29
removing reference to cdn_name parameter

### LINKED ISSUE #1180: Clean up docs for CDN_name parameter
Now that it is a table, the references to the parameter should be cleaned up:

http://traffic-control-cdn.net/docs/latest/admin/traffic_router.html?highlight=cdn_name

jheitz200:
fixes #1180


dneuman64:
Looks like there is one more here:  http://traffic-control-cdn.net/docs/latest/development/traffic_ops_api/v12/cache_stats.html?highlight=cdn_name

PR #1572, opened 2016-06-30
More purge cleanup. Fixes #1571

### LINKED ISSUE #1571: Purge TTL  parameters don't make sense and don't work.
ttl_min_hours ttl_max_hours should be replaced by maxRevalDurationDays and a hard 1 hour minimum.

knutsel:
Jeremy: Please review. I need to still test the API case, so don't merge yet, but wanted you to get eyes on this, I know it is late....


traffic_ops/app/lib/API/Job.pm
```diff
@@ -205,10 +205,10 @@ sub is_valid_date_format {
 sub is_ttl_in_range {
 	my $self  = shift;
 	my $value = shift;
-	my $min_hours =
-		$self->db->resultset('Parameter')->search( { name => "ttl_min_hours" }, { config_file => "regex_revalidate.config" } )->get_column('value')->first;
-	my $max_hours =
-		$self->db->resultset('Parameter')->search( { name => "ttl_max_hours" }, { config_file => "regex_revalidate.config" } )->get_column('value')->first;
+	my $min_hours = l;
```
mitchell852: is that a 1? looks like an "el"

traffic_ops/app/lib/API/Job.pm
```diff
@@ -205,10 +205,10 @@ sub is_valid_date_format {
...
+	my $max_days =
```
mitchell852: what if the "maxRevalDurationDays" param doesn't exist? this will blow up. some kind of default you can fall back to?

mitchell852: actually, i suppose it's fine. you seed the database with this param.

mitchell852:
ooops, you got a couple broken tests:

`prove -qrp t`

mitchell852:
I tested the API. Jobs between 1 and 2160 TTL (based on maxRevalDurationDays=90) inserted a record in Job table and created a line entry in regex_revalidate.config

PR #1575, opened 2016-06-30
Update TO default profiles for 1.6

mitchell852:


traffic_ops/install/data/profiles/profile.global.traffic_ops
```diff
@@ -47,7 +47,7 @@
         }
     ],
     "profile": {
-        "description": "Global Traffic Ops profile, DO NOT DELETE",
-        "name": "GLOBAL"
+        "description": "Global profile",
+        "name": "Global"
```
robert-butts: The names and descriptions of all these profiles are whatever you gave the script. Can you set them back to the default values?

You can run the script again with `-name` and `-description` set to the original value, or just manually revert the JSON, whichever is easier.

PR #1577, opened 2016-07-01
Traffic Router now can redirect with https using keystore

### LINKED ISSUE #1023: Enable HTTPS for Content Routing Type = HTTP
Would like to enable HTTPS for delivery services using HTTP 302 redirects, as is currently possible when using DNS as the Content Routing type.

### LINKED ISSUE #1457: Option for automatic redirection of http to https from edge->client
This came up recently with a new delivery service -- the need is for having the option of automatically redirecting customers to https before beginning proxy communication.

### LINKED ISSUE #1505: Traffic Router HTTP Redirection to HTTPS
Discussions as started on adding HTTPS to Traffic Router and we have a specific use case for Video Delivery service over HTTPS. This could be either enhancement or question.

The need would be to use HTTP and HTTPS as a request method to Traffic Router, but the redirection would always end up being an HTTPS location. Since this is for video delivery, we want to keep the Request Routing to minimal delay, keeping resource usage on the Client/Server low.

Seems like this should be the default behavior. 

Somewhat related to #1457 although this is Edge server requirement, which this here is kind of a similar ask except done at the Traffic Router.

trevorackerman:
fixes #1023 
fixes #1457 
fixes #1505 

The next step is to have Traffic Router poll traffic ops and import certificates via KeyStoreHelper.importCertificate()


jianli123:
Do we have the issue number to track the remaining "The next step is to have Traffic Router poll traffic ops and import certificates via KeyStoreHelper.importCertificate()"?  thanks

trevorackerman:
I don't have an issue number to track what you're speaking about. I should have another PR over the next several days that will cover this.

Issue #1619, opened 2016-07-21
Proposed server table changes


dg4prez:
1) Maximum hostname, domain name length - standards require these to be no longer than 63 bytes per segment.  Currently they're 45 char - suggest increasing to 63 for hostname, and 63 at minimum for domain name.  While domain name could technically be longer, i'd want some sort of safeguard built in to count per segment built into the ui before changing it.

2) Remove the unique requirement for host_name, ip_address, and ip6_address.  As long as we don't have identical profiles and cachegroups, we can still tell the difference between FQDN and we need a way to put the same servers into multiple groups for some MSO tricks, at least in the short term.  We may want to consider displaying the full FQDN in the server assignment list rather than just the host name.  This is also needed for host name because we've had requests to setup origins with the same hostname and different domain names.

3) Allow null values for interface name, ip_netmask, and ip_gateway.  When describing servers that we do not control such as origins, there's no need for this data and we end up fudging it.  Suggest changing the UI to require those fields when needed and allow the user not to enter them when they're not.  Relying on the database to block "bad" entries is bad design anyway and creates ugly errors that can't be recovered from without starting over.

knutsel:
@limited : I remember you had the same problem with the uniqueness of ip address? Do I remember right? Does this fix your use case, or did you have something else in mind?

limited:
Our use case was putting the same origin servers in two different CDNs. I think the above suggestion 2 would take care of our issue. We did discuss allowing dupe {hostname, ip_address, ip6_address} but with some type of popup warning box to the user. If thats not a ton of work I think it could still be useful.

dg4prez:
if we could get the DB changed now and make the UI changes going forward, that would be terribly useful.

dg4prez:
https://github.com/Comcast/traffic_control/pull/1623

jeffmart:
I will second option number 2. We performed this in the DB, had done some initial testing. This is currently not in the code but given this is being brought up from another source is this something that we should re-review and consider? Note: right now we are using a VIP to work around this.

dneuman64:
this was fixed by #1623

mitchell852:
update on this:

1. done

2. unique constraint was removed from hostname but ip_address and ip6_address must be unique per profile.

3. not done due to the problems this would cause with edge and mid servers for example.

PR #1623, opened 2016-07-21
modify server table to allow duplicate entries, increase char length

dg4prez:
This will allow duplicate IP and hostname entries, as well as increase the character limit from 45 to 63, which is the max allowed per segment according to rfc.  This also removes the DB requirement to have subnet mask, port name, and gateway.  The UI still requires this - items like ORG should be modified to not require these entries in the UI.


dg4prez:
https://github.com/Comcast/traffic_control/issues/1619

mitchell852:
if you change your previous comment to 'fixes #issueNumber', the issue will be closed when the PR is merged...if that's what you want...

mitchell852:
Removing constraints from the TO database (i.e. a unique constraint or a not null constraint) could potentially harm the integrity of the data upon which the CDN relies. To me, it seems like without the unique constraint, you could create a cachegroup with servers in it like this:

server A: 2.2.2.2
server B: 2.2.2.2
server C: 2.2.2.2

which may not break the CDN per se but will cause it to perform at a suboptimal level (load balancing is basically turned off). Maybe there are other implications that I am not aware of. I just want to make sure there are no unwanted side effects of this change. Anyone else have any thoughts or concerns?

mitchell852:
so a couple of things off the top of my head needs to be done in addition to the database schema changes:

1. it sounds like UI modification is needed to not require subnet mask, port name, and gateway if server.type == ORG on server create/update
2. API modification is needed to not require subnet mask, port name, and gateway if server.type == ORG on server create/update
3. test cases
a. testcase to test that duplicate IP and hostname entries are accepted and persisted to the db
b. testcast to test # 2 (create or update an ORG server thru the api without subnet mask, port name, or gateway)

dg4prez:
Are there any issues with keeping the duplicate hostnames change?  We need either that or to merge the hostname into a single FQDN.  We have encountered cases where origins have identical hostnames but different domains, especially in cases where we are setting up for dual-origin configs.

mitchell852:
It makes sense to me that server.host_name should not be unique as you may have 2 origin servers that look like:

foo-bar.domain1.com
foo-bar.domain2.com

how about instead of dropping the unique constraints altogether, add multiple fields to the constraints like this maybe (i think this would work but i'm no DB expert):

```alter table server drop index host_name;

ALTER TABLE `server` ADD UNIQUE `unique_index`(`host_name`, `domain_name`, `cdn_id`);

alter table server drop index cs_ip_address_UNIQUE;

ALTER TABLE `server` ADD UNIQUE `unique_index`(`ip_address`, `cdn_id`);

alter table server drop index ip6_address;

ALTER TABLE `server` ADD UNIQUE `unique_index`(`ip6_address`, `cdn_id`);```

**Comment by mitchell852** [comment]
also, you'll want to check that these indexes actually exist before attempting to drop them. dropping an non-existent index will cause the migration to fail.

**Comment by mitchell852** [comment]
i made some changes to the migration file, however, the UI and presumedly the API still enforces interface_name, ip_netmask and ip_gateway as required fields. i'll hold off on merging this until that is done. i can chip in on that next week.

File: docs/source/development/traffic_ops_api/v12/server.rst [hunk 1]
```diff
@@ -661,13 +661,13 @@ Server
   +------------------+----------+------------------------------------------------+
   | cachegroup       | yes      | cache group name                               |
   +------------------+----------+------------------------------------------------+
```
mitchell852: change these back to required

traffic_ops/app/db/migrations/20160826111234_modify_server_table.sql
```diff
@@ -0,0 +1,65 @@
...
+PREPARE stmt2 FROM @sqlstmt2;
+EXECUTE stmt2;
+
+set @exist3 := (select count(*) from information_schema.statistics where table_name = 'server' and index_name = 'host_name' and table_schema = database());
+set @sqlstmt3 := if( @exist3 > 0, 'alter table server drop index host_name', 'alter table server');
+PREPARE stmt3 FROM @sqlstmt3;
+EXECUTE stmt3;
+
+alter table server modify host_name varchar(63) not null;
+alter table server modify domain_name varchar(63) not null;
```
mitchell852: change interface_name, ip_netmask and ip_netmask back to not null

traffic_ops/app/lib/API/Server.pm
```diff
@@ -412,18 +413,9 @@ sub check_server_params {
 	if ( !defined( $json->{'cachegroup'} ) ) {
 		$errFields{'cachegroup'} = 'is required';
 	}
```
mitchell852: put these lines back in

PR #1664, opened 2016-08-02
Allow origin on port different than default 80

### LINKED ISSUE #1203: TO: parent.config generation needs to separate origin port from domain
This is a specific case I'm running into. I want to use the `HTTP_LIVE` delivery service type along with an origin which has a different port number than 80. In this case I used the IP address, but it doesn't matter.

Origin URL : `http://192.168.1.1:8080`

The generated parent.config looks like this : `dest_domain=192.168.1.1:8080 go_direct=true`

The proper end config : `dest_domain=192.168.1.1 port=8080 go_direct=true`

smalenfant:
We have now used this in both lab and production. It adds the `port=80` in the parent.config line for the specific origins. The `dest_domain=.` is untouched for edge. 

I haven't tested the primary/secondary yet as there is possibly other issues associated with that.

Edge Example : 

```
dest_domain=tld1.net port=80 parent="mid1:80|0.999;mid2:80|0.999;mid3:80|0.999"  round_robin=consistent_hash go_direct=false qstring=ignore
dest_domain=tld2.net port=80 go_direct=true
```

Mid Example :

```
dest_domain=tld1.net port=80 parent="origin1:80|1.0;origin2:80|1.0;origin3:80|1.0" round_robin=consistent_hash go_direct=false parent_is_proxy=false
```

Requesting to be put in 1.7.0, but this might be a major change. 

Fixes #1203


smalenfant:
Questioning this change here as I don't use rank_parents... Seems like it should work, unless there is an origin URI with port information. Probably needs changed to `$pinfo->{$org_uri}`

```
				my @ranked_parents = ();
 -				if ( exists( $pinfo->{$org_fqdn} ) ) {
 -					@ranked_parents = sort by_parent_rank @{ $pinfo->{$org_fqdn} };
 +				if ( exists( $pinfo->{$org_uri->host} ) ) {
 +					@ranked_parents = sort by_parent_rank @{ $pinfo->{$org_uri->host} };
  				}
```

smalenfant:
MSO will not rank and put the parent information if the origin server has a Port in the URL. Although I don't think MSO has been tested with origin with different origin port. I'd say the PR still good as is.

NOTE: Don't put a URL with Port 80 like this : `http://example.net:80` as the parent will not show up in the config. The following configuration will be written :

```
dest_domain=example.net port=80 parent="" round_robin=consistent_hash go_direct=false parent_is_proxy=false
```

smalenfant:
Closing in favor of #1783

PR #1672, opened 2016-08-03
Traffic Portal: build without using ant

### LINKED ISSUE #1666: Traffic Portal: build rpm depends on ant
dependency on ant makes it harder to use common build functions (naming rpm, etc..).   This should be removed.

dangogh:
no ant required;  builds using same structure as other rpms.

Fixes #1666


mitchell852:
tested. works with centos-6.7...with try centos 7 now.

mitchell852:
works with centos7 too. nice work. merging...

PR #1687, opened 2016-08-09
TR now sends 503 to http clients accessing delivery services only supporting https traffic

trevorackerman:
Only clients connecting via https will receive a 302


trevorackerman:
getting rid of this branch, it'll be easier to re-do this in a new branch

PR #1690, opened 2016-08-09
Traffic Stats now correctly calculates daily stats

### LINKED ISSUE #1669: Daily Bytesserved Data is incorrect
In systems with lots of cache servers (> 1000), the Daily Bytesserved data is not being generated correctly.  It is often times much, much lower than it should be.  My hunch is that it has something to do with the amount of data points being returned in the query.  Maybe optimizing the query (or using continuous query data) will resolve the issue.

dneuman64:
Traffic Stats now uses continuous queries to calculate daily stats.  This is to address an issue where daily stats, specifically daily_bytesserved, was not correctly calculated once our CDN got to be > 1000 caches.  

This fixes #1669

### DIFF
--- a/traffic_stats/traffic_stats.go
+++ b/traffic_stats/traffic_stats.go
@@ -48,7 +48,7 @@ const (
 
 const (
 	defaultPollingInterval             = 10
-	defaultDailySummaryPollingInterval = 60
+	defaultDailySummaryPollingInterval = 30
 	defaultConfigInterval              = 300
 	defaultPublishingInterval          = 30
 	defaultMaxPublishSize              = 10000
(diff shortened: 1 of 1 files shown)


traffic_stats/traffic_stats.go
```diff
@@ -259,97 +259,123 @@ func calcDailySummary(now time.Time, config StartupConfig, runningConfig Running
...
+			for _, record := range row.Values {
+				t := record[0].(string)
+				if record[1] != nil {
+					cdn := record[1].(string)
+					value, err := record[2].(json.Number).Float64()
+					if err != nil {
+						log.Errorf("Couldn't parse value from record %v\n", record)
+						continue
+					}
+					value = value / 1000000
```
robert-butts: What is 1000000? If we put it in a variable, e.g. `kilobytesInMegabyte := 1000000` then `value / kilobytesInMegabyte` would be more obvious what it's doing

traffic_stats/traffic_stats.go
```diff
@@ -259,97 +259,123 @@ func calcDailySummary(now time.Time, config StartupConfig, runningConfig Running
...
+					bp.AddPoint(pt)
+				}
+			}
+		}
+	}
+	config.BpsChan <- bp
+}
+
+func calcDailyBytesServed(client influx.Client, bp influx.BatchPoints, startTime time.Time, endTime time.Time, config StartupConfig) {
+	queryString := fmt.Sprintf("select mean(value) from \"monthly\".\"bandwidth.cdn.1min\" where time > '%s' and time < '%s' group by time(1m), cdn", startTime.Format(time.RFC3339), endTime.Format(time.RFC3339))
```
robert-butts: If this used ` quotes, it could avoid escaping the "

traffic_stats/traffic_stats.go
```diff
@@ -259,97 +259,123 @@ func calcDailySummary(now time.Time, config StartupConfig, runningConfig Running
...
-			pt, err = influx.NewPoint(
-				statsSummary.StatName,
+				"daily_bytesserved",
 				tags,
 				fields,
 				startTime,
 			)
 			if err != nil {
-				errHndlr(err, ERROR)
+				fmt.Printf("error adding creating data point for max Gbps...%v\n", err)
```
robert-butts: Should this be `log.Errorf` instead of `fmt.Printf`?

traffic_stats/traffic_stats.go
```diff
@@ -259,97 +259,123 @@ func calcDailySummary(now time.Time, config StartupConfig, runningConfig Running
...
+						"value": value,
+					}
+					pt, err := influx.NewPoint(
+						"daily_maxgbps",
+						tags,
+						fields,
+						statTime,
+					)
+					if err != nil {
+						fmt.Printf("error adding creating data point for max Gbps...%v\n", err)
```
robert-butts: Should this be log.Errorf instead of fmt.Printf?

PR #1696, opened 2016-08-10
TO now allows editing delivery services in different cdn's with the same host regex

trevorackerman:
Used to get a bogus error that I was trying to add a host regex that was already taken by another delivery service. However the other delivery service was not in the same cdn.

### DIFF
--- a/traffic_ops/app/lib/UI/DeliveryService.pm
+++ b/traffic_ops/app/lib/UI/DeliveryService.pm
@@ -51,7 +51,7 @@ sub edit {
 	my $data = $rs_ds->single;
 	my $action;
 	my $regexp_set = &get_regexp_set( $self, $id );
-	my $cdn_domain = &get_cdn_domain( $self, $id );
+	my $cdn_domain = $self->get_cdn_domain_by_ds_id($id);
 	my @example_urls = &get_example_urls( $self, $id, $regexp_set, $data, $cdn_domain, $data->protocol );
 
 	my $server_count = $self->db->resultset('DeliveryserviceServer')->search( { deliveryservice => $id } )->count();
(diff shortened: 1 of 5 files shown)


mitchell852:
i'm gonna test this PR. can you create an issue for this bug?

traffic_ops/app/lib/UI/DeliveryService.pm
```diff
@@ -386,33 +386,46 @@ sub check_deliveryservice_input {
...
-						? $self->param( 're_re_' . $order_no ) . $cdn_domain
-						: $self->param( 're_re_new_' . $order_no ) . $cdn_domain;
-					my $new_order =
-						defined( $self->param( 're_order_' . $order_no ) )
-						? $self->param( 're_order_' . $order_no )
-						: $self->param( 're_order_new_' . $order_no );
-					my $rs =
-						$self->db->resultset('DeliveryserviceRegex')->search( undef, { prefetch => [ { regex => undef }, { deliveryservice => undef } ] } );
+
+				my $is_re_type_host_regex = defined($self->param('re_type_' . $order_no)) && $self->param('re_type_' . $order_no) eq 'HOST_REGEXP';
```
mitchell852: This logic needs to be applied to the API namespace as well or the behavior of the UI != API. I'm guessing this logic needs to be applied to https://github.com/Comcast/traffic_control/blob/master/traffic_ops/app/lib/API/DeliveryService.pm#L572

Also, once  you update the API, can you write a test? Maybe test that 2 identical regex's from the same cdn fail and that 2 identical regex's from different cdn's pass?

mitchell852:
I suspect this test failure is a result of this change. Can you check it out, Trevor?

`#   Failed test '200 OK'
#   at t/api/1.2/deliveryservice.t line 251.
#          got: '400'
#     expected: '200'
# {"alerts":[{"level":"error","text":"Another delivery service is already using host regex .*\\.ds_1\\..*foo.com"}]}`

Issue #1703, opened 2016-08-16
non-admin user should can't delete the secure parameter


jifyang:
The secure parameter can't be create/updated/viewed by non-admin user. It also should can't be deleted by non-admin user. The operation validation should be added for "delete".

PR #1704, opened 2016-08-16
non-admin user should can't delete the secure parameter

jifyang:
fix the issue #1703

### DIFF
--- a/traffic_ops/app/lib/UI/Parameter.pm
+++ b/traffic_ops/app/lib/UI/Parameter.pm
@@ -151,7 +151,7 @@ sub delete {
 	else {
 		my $secure = $self->db->resultset('Parameter')->search( { id => $id } )->get_column('secure')->single();
 		if ( (1==$secure) && !&is_admin($self) ) {
-			$self->flash( alertmsg => "No can do for secure parameter. Get more privs." );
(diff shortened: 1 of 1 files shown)


traffic_ops/app/lib/UI/Parameter.pm
```diff
@@ -149,10 +149,16 @@ sub delete {
 		$self->flash( alertmsg => "No can do. Get more privs." );
 	}
 	else {
-		my $p_name = $self->db->resultset('Parameter')->search( { id => $id } )->get_column('name')->single();
-		my $delete = $self->db->resultset('Parameter')->search( { id => $id } );
-		$delete->delete();
-		&log( $self, "Delete parameter " . $p_name, "UICHANGE" );
+		my $secure = $self->db->resultset('Parameter')->search( { id => $id } )->get_column('secure')->single();
+		if ( (1==$secure) && !&is_admin($self) ) {
+			$self->flash( alertmsg => "No can do for secure parameter. Get more privs." );
```
mitchell852: Can you make this error message more clear like "Forbidden. Admin role required to delete a secure parameter". thanks.

jifyang: Updated. Thanks.

PR #1708, opened 2016-08-17
make sure to pass in ds id for check_params on DS put api

trevorackerman:


mitchell852:
unit tests pass now. merging...

PR #1726, opened 2016-08-19
TR now will not apply a new cr config if it has https delivery services without certs

trevorackerman:


traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/config/CertificateChecker.java
```diff
@@ -0,0 +1,53 @@
+package com.comcast.cdn.traffic_control.traffic_router.core.config;
+
+import com.comcast.cdn.traffic_control.traffic_router.keystore.KeyStoreHelper;
+import org.apache.log4j.Logger;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+public class CertificateChecker {
+	private final static Logger LOGGER = Logger.getLogger(CertificateChecker.class);
+	private final KeyStoreHelper keyStoreHelper = KeyStoreHelper.getInstance();
```
robert-butts: This must be mutated by another thread, if it's possible for certificatesAreValid() in ConfigHandler.java:135 to become true, right?

Then, either `KeyStoreHelper.KeyStore` or `KeyStoreHelper` needs declared volatile, right?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/secure/CertificatesClient.java
```diff
@@ -36,18 +47,18 @@ public void refreshData() {
 		persistCertificates(certificateDataList);
 	}
 
-	public String fetchRawData() {
+	public int fetchRawData(final StringBuilder stringBuilder) {
```
robert-butts: A comment indicating the return value would be nice here. You can't really tell what the int is without reading the full function.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/secure/CertificateData.java
```diff
@@ -27,24 +30,34 @@ public void setCertificate(final Certificate certificate) {
...
+	}
+
+	public void setHostname(final String hostname) {
+		this.hostname = hostname;
+	}
+
 	@SuppressWarnings("PMD")
-	public boolean equals(final Object o) {
+	@Override
+	public boolean equals(Object o) {
```
robert-butts: Why was this changed to not be final? The `equals` function shouldn't modify it, right?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/secure/CertificateData.java
```diff
@@ -27,24 +30,34 @@ public void setCertificate(final Certificate certificate) {
...
 		if (this == o) return true;
 		if (o == null || getClass() != o.getClass()) return false;
 
-		final CertificateData that = (CertificateData) o;
+		CertificateData that = (CertificateData) o;
 
 		if (deliveryservice != null ? !deliveryservice.equals(that.deliveryservice) : that.deliveryservice != null)
 			return false;
-		return certificate != null ? certificate.equals(that.certificate) : that.certificate == null;
+		if (certificate != null ? !certificate.equals(that.certificate) : that.certificate != null) return false;
```
robert-butts: I had to stare at this for like 45 seconds to figure out what it does in every case. If it were two separate `if`s, it'd take, like, 10% the time to mentally parse.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/secure/CertificateData.java
trevorackerman: I updated the class with a field told IntelliJ to regenerate the equals method.
That caused the final keyword to be dropped unintentionally, I usually add it back.

I always use IntelliJ's generation of equals and hashCode, it doesn't read well but it always works and is performant.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/secure/CertificateData.java
trevorackerman: see my comment above

PR #1727, opened 2016-08-19
Fix Traffic Monitor 2.0 Events, Fix Kbps

robert-butts:
There were two problems: (1) the EventsThreadsafe type wasn't using pointers, which resulted in set values not propogating, and (2) the GUI was iterating forward, and skipping indices older than the last, but the events array is backwards, so it needs to iterate backwards.

This fixes both.


traffic_monitor/experimental/traffic_monitor/cache/cache.go
```diff
@@ -150,19 +160,40 @@ func (handler Handler) Handle(id string, r io.Reader, err error, pollId uint64,
 	fmt.Printf("DEBUG poll %v %v handle end\n", pollId, time.Now())
 }
 
+// outBytes takes the proc.net.dev string, and the interface name, and returns the bytes field
+// \todo
```
dneuman64: huh?

robert-butts: You heard me.

PR #1729, opened 2016-08-20
Traffic Ops - Set HTTPS port for a server

dneuman64:
Traffic Ops now supports setting an HTTPS port for a server.  This is related to HTTPS support for HTTP deliveryservices.  Traffic Router uses this value to determine which port to use when re-directing HTTPS requests.  If no port is set, TR will use a default port of 443.
A new column named https_port has been added to the server table, UI, and API.  This column is set to 443 by seeds.sql.  

Sorry for the extra formatting on stuff that didn't change.


traffic_ops/app/lib/API/Server.pm
```diff
@@ -617,6 +626,9 @@ sub check_server_params {
 	if ( defined( $json->{'tcpPort'} ) && $json->{'tcpPort'} !~ /\d+/ ) {
 		return ( \%params, $json->{'tcpPort'} . " is not a valid tcp port" );
 	}
+	if ( defined( $json->{'httpsPort'} ) && $json->{'httpsPort'} !~ /\d+/ ) {
+		return ( \%params, $json->{'httpsPort'} . " is not a valid tcp port" );
```
mitchell852: Not sure if it's a big deal but I can set an https port # greater than 65536 thru the UI and API.

dneuman64: I thought about that and then just decided to follow the logic that was already there.

mitchell852: sounds good

traffic_ops/app/lib/API/Server.pm
```diff
@@ -447,6 +450,12 @@ sub check_server_params {
 	elsif ( !defined($update_base) ) {
 		$params{'tcpPort'} = 80;
 	}
+	if ( defined( $json->{'httpsPort'} ) ) {
+		$params{'httpsPort'} = int( $json->{'httpsPort'} );
+	}
+	elsif ( !defined($update_base) ) {
```
mitchell852: I would drop this elseif because since the field is not required in the request json, someone may exclude the field and inadvertently reset the value to 443.

mitchell852: oh wait, nm. i didn't read the elseif properly. i guess it checks on $update_base (whatever that is)

traffic_ops/app/lib/API/Server.pm
```diff
@@ -817,6 +832,7 @@ sub update {
 				host_name      => defined( $params->{'hostName'} )      ? $params->{'hostName'}      : $update->host_name,
 				domain_name    => defined( $params->{'domainName'} )    ? $params->{'domainName'}    : $update->domain_name,
 				tcp_port       => defined( $params->{'tcpPort'} )       ? $params->{'tcpPort'}       : $update->tcp_port,
+				tcp_port       => defined( $params->{'httpsPort'} )       ? $params->{'httpsPort'}       : $update->https_port,
```
mitchell852: ruh roh

dneuman64: oops, good catch!

traffic_ops/app/lib/Schema/Result/Server.pm
```diff
@@ -362,6 +369,12 @@ __PACKAGE__->add_unique_constraint("host_name", ["host_name"]);
 
 __PACKAGE__->add_unique_constraint("ip6_address", ["ip6_address"]);
```
mitchell852: i think 372-377 needs to be removed

PR #1763, opened 2016-08-29
TO now returns all certs for sslkeys instead of just first 10

trevorackerman:
solr search behind riak only returns the first 10 by default and they are not consistently ordered.


dneuman64:
Good catch.

dg4prez:
remap.config now shows an https entry.

PR #1764, opened 2016-08-29
TO now shows 'manage ssl keys' for http delivery service protocol 'http-to-https'

trevorackerman:


dg4prez:
Testing this in the lab, I'm not seeing a remap.config entry when a service is switched from http/https to http to https.

PR #1766, opened 2016-08-30
Centos7 ort change

jrushford:
The os_version() sub in traffic_ops_ort.pl needs to be more robust when parsing the uname -r output.


dneuman64:
Verified on Centos6.5 and Centos7.2 👍

Issue #1584, opened 2016-07-06
need a way to configure qstring value in parent.config by DeliveryService


dneuman64:
Currently the qstring value in the parent.config is based on a parameter on the server profile.  We need to make this configurable so that the qstring can be set to "consider" if we want the qstring to be used when hashing to a parent.

smalenfant:
The proposal here would be to set the parent.config line to `qstring=consider` when the Query String Handling == 0 (use qstring in cache key, and pass up). 

I got a patch I'm testing in my lab.

dneuman64:
yeah, that is the same idea I had.  I dont know if it is perfect but it is good enough

Issue #1478, opened 2016-06-01
TO: MID parent.config generation takes longer than expected with Multi-Site


smalenfant:
Was testing some scripts and API calls and found out the MID `parent.config` generation is taking longer than expected. Would need some help to verify if this is reproducible on another system.

Seems like it takes about 200ms processing time for each Delivery Service which are Multi-Site.

Traffic Ops version 1.5.1

```
catl0w10ct16662:traffic-control-python smalenfa$ time python ort.py 
cdn1cdmid0001 3971 ms
cdn1cdmid0002 4004 ms
cdn1cdmid0003 3043 ms
cdn1cdmid0004 3045 ms
cdn1cdmid0005 3036 ms
cdn1cdmid0006 3096 ms
psp6cdmid01 3307 ms
psp6cdmid02 3469 ms
psp6cdmid03 3509 ms
psp6cdmid04 3546 ms
```

Traffic Ops access.log :
```
::ffff:24.248.74.254 - - [01/Jun/2016:03:11:15 +0000] "GET /genfiles/view/cdn1cdmid0001/parent.config HTTP/1.1" 200 480 2950971 "python-requests/2.7.0 CPython/2.7.11 Darwin/15.5.0"
::ffff:24.248.74.254 - - [01/Jun/2016:03:11:18 +0000] "GET /genfiles/view/cdn1cdmid0002/parent.config HTTP/1.1" 200 481 2905609 "python-requests/2.7.0 CPython/2.7.11 Darwin/15.5.0"
::ffff:24.248.74.254 - - [01/Jun/2016:03:11:21 +0000] "GET /genfiles/view/cdn1cdmid0003/parent.config HTTP/1.1" 200 481 2981979 "python-requests/2.7.0 CPython/2.7.11 Darwin/15.5.0"
::ffff:24.248.74.254 - - [01/Jun/2016:03:11:24 +0000] "GET /genfiles/view/cdn1cdmid0004/parent.config HTTP/1.1" 200 481 2940920 "python-requests/2.7.0 CPython/2.7.11 Darwin/15.5.0"
::ffff:24.248.74.254 - - [01/Jun/2016:03:11:27 +0000] "GET /genfiles/view/cdn1cdmid0005/parent.config HTTP/1.1" 200 481 2936758 "python-requests/2.7.0 CPython/2.7.11 Darwin/15.5.0"
[...]

Issue #1278, opened 2016-04-06
TO: parent.config in multi-site origin shows duplicate lines


mtorluemke:
parent.config has entries like this:

dest_domain=origin1.origin.com parent="1.2.3.4:80|1.0;" round_robin=consistent_hash go_direct=false parent_is_proxy=false
dest_domain=origin1.origin.com parent="1.2.3.4:80|1.0;" round_robin=consistent_hash go_direct=false parent_is_proxy=false

PR #1783, opened 2016-08-31
Multiple configuration file generation issues

smalenfant:
Commit contains all the issues attempting to fix. Please review carefully. There is enhancement and performance fixes (like MSO config).

This has been tested with a testbed similar to our production system, but please verify with your use cases.

Notes: 

- Before applying to a production system, make sure the delivery service has proper configuration for Query String Handling. Currently, the change only affects DS which has the QSH set to `use string in cache key, and pass up`.
- Origin Port. This has been running in production for a while, although it doesn't cover the case where the origin port is set on the URL. Fox example, don't set `http://example.com:80` as it won't work. Although I'm about 99% sure this didn't work either before. I'll be looking in fixing this in the future.


mitchell852:
does this map back to any issues that are open?

mitchell852:
oh wait. i can see in the commits this addresses #1478, #1203, #1278 and #1584

dg4prez:
Tested and confirmed: these fixes appear to work as expected.

mitchell852:
checked that duplicate parent.config entries are not created (#1278), qstring is set appropriately to "consider" (where qstring_ignore=0) (#1584), dest_domain and port are set in parent.conf appropriately based on ds.org_server_fqdn (port=80 if no port specified on ds.org_server_fqdn and port=x if port is specified) (#1203) and parent.config is generated in a reasonable timeframe (#1478). 

@dg4prez also tested and @dangogh blessed it with his Perl wand. once this is installed in our lab environment, regression testing will be required to ensure ATS functionality hasn't been affected in a negative way. merging...

PR #1798, opened 2016-09-02
Traffic Router now always cleans up neustar downloads and closes stream resources, fixes #1782

### LINKED ISSUE #1782: Traffic Router not removing files correctly and leaving open file handles
Traffic Router is not deleting files from /opt/tomcat/temp correctly causing there to be open file handles which show (deleted).   This could cause disk to fill up and tomcat would need to be restarted to free up the disk space.

A sample from a lab TR:

```
[root@my-ccr-01 ~]# lsof | grep deleted
java      30206        root   53w      REG                8,3          0     155686 /opt/apache-tomcat-6.0.33/temp/loc5328418763434655085.dat (deleted)
java      30206        root   55w      REG                8,3          0     155700 /opt/apache-tomcat-6.0.33/temp/loc7395668089700974662.dat (deleted)
java      30206        root   57w      REG                8,3          0     155702 /opt/apache-tomcat-6.0.33/temp/loc4173128634880722976.dat (deleted)
java      30206        root   59w      REG                8,3          0     155703 /opt/apache-tomcat-6.0.33/temp/loc6329917757855660469.dat (deleted)
```

trevorackerman:
fixes #1782

### DIFF
--- a/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/dns/DNSKeyPairWrapper.java
+++ b/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/dns/DNSKeyPairWrapper.java
@@ -47,20 +47,20 @@ public class DNSKeyPairWrapper extends DnsKeyPair {
 		this.expiration = new Date(1000L * keyPair.getLong("expirationDate"));
 		this.ttl = keyPair.optLong("ttl", defaultTTL);
 		this.name = keyPair.getString("name");
-		//this.status = keyPair.getString("status"); // this field is used by Traffic Ops; we detect expiration by using the above dates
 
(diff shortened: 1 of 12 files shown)


elsloo:
Perhaps I'm missing something, but I don't see where we're closing files that may be open prior to deleting them. That's what causes the orphaned file handles that are only reclaimed once the program exits.

Also, it seems as though these changes only apply to Neustar when we should be checking for the same issue with any downloaded resource that we might swap in favor of a newer version (MaxMind, CZF, etc).

trevorackerman:
The following is the root source of the defect we observed. Notice that we're not closing the stream.

https://github.com/Comcast/traffic_control/blob/master/traffic_router/neustar/src/main/java/com/comcast/cdn/traffic_control/traffic_router/neustar/data/TarExtractor.java#L44

I searched through the code and found a handful more places that although unlikely to cause a problem needed attention for closing resources. In most cases they've been converted to Java 7's try-with-resource convention.

PR #1801, opened 2016-09-06
remove dead cpan repo ref

dangogh:
cpan.yimg.com apparently no longer exists.  Removed from docker responses..


dangogh:
Fixes #1802

JJ:
This is basically done, right?

PR #1807, opened 2016-09-06
ISSUE: 1806 traffic_ops.ort.pl fails with syntax error

jrushford:
fixed traffic_ops_ort.pl


dneuman64:
Going to fix this inline

PR #1843, opened 2016-09-15
Add NetAcuityGeolocationService

bengunter:
@trevorackerman I apologize for getting your name wrong in the commit message.


trevorackerman:
The default build will have to not depend on the net acuity artifact unless it can be distributed under one of these licenses - http://www.apache.org/legal/resolved.html#category-a

You'll need to add a maven profile so that building with the net acuity artifact is optional.

Follow the example of what we did for using the neustar version of the geolocation services as a guide. I'd recommend making a module like the neustar module to accomplish this.

I have 2 high priority items I must tackle first, and after I get those done I'd be happy to help guide you by forking your copy of the project and making a pull request against it. But it may be about a week or more before I can do that.

dneuman64:
We have moved to a new repo.  If this functionality is desired, please submit a PR there.  https://github.com/apache/incubator-trafficcontrol

Issue #1851, opened 2016-09-16
TR throws an error on startup when trying to download Geo Database


dneuman64:
If the Geo Database already exists and TR is restarted, the following error is thrown in the log:  

> 2016-09-15 15:50:29,693 [FATAL] traffic_router.core.loc.AbstractServiceUpdater: [GeolocationDatabaseUpdater] Caught exception while attempting to download: https://trafops.kabletown.net/MaxMind/auto/GeoIP2-City.mmdb.gz
> java.net.SocketException: java.security.NoSuchAlgorithmException: Error constructing implementation (algorithm: Default, provider: SunJSSE, class: sun.security.ssl.SSLContextImpl$DefaultSSLContext)
> 	at javax.net.ssl.DefaultSSLSocketFactory.throwException(SSLSocketFactory.java:248)
> 	at javax.net.ssl.DefaultSSLSocketFactory.createSocket(SSLSocketFactory.java:255)
> 	at sun.net.www.protocol.https.HttpsClient.createSocket(HttpsClient.java:409)
> 	at sun.net.NetworkClient.doConnect(NetworkClient.java:162)
> 	at sun.net.www.http.HttpClient.openServer(HttpClient.java:432)
> 	at sun.net.www.http.HttpClient.openServer(HttpClient.java:527)
> 	at sun.net.www.protocol.https.HttpsClient.<init>(HttpsClient.java:275)
> 	at sun.net.www.protocol.https.HttpsClient.New(HttpsClient.java:371)
> 	at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.getNewHttpClient(AbstractDelegateHttpsURLConnection.java:191)
> 	at sun.net.www.protocol.http.HttpURLConnection.plainConnect0(HttpURLConnection.java:1104)
> 	at sun.net.www.protocol.http.HttpURLConnection.plainConnect(HttpURLConnection.java:998)
> 	at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect(AbstractDelegateHttpsURLConnection.java:177)
> 	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1512)
> 	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1440)
> 	at sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:254)
[...]

PR #1866, opened 2016-09-20
Traffic Router now accepts PKCS1 formatted certificates for https

### LINKED ISSUE #1850: Traffic Router should support pkcs1 and pkcs8 encoded private keys for SSL
Currently, TR only support PKCS8 encoded private keys.  Since TO creates private keys that are PKCS1 encoded TR should support both. 
When TR tries to load the keys that are currently created by TO is throws the following error:  `java.security.spec.invalidkeyspecexception: java.security.invalidkeyexception: ioexception : algid parse error, not a sequence `

trevorackerman:
Fixes #1850


dneuman64:
It looks like we now ONLY support PKCS1 formatted private keys.  Is that the intention?

trevorackerman:
Why do you believe it only supports PKCS1?

dneuman64:
Looks like the cert eventually did get loaded

trevorackerman:
Thanks for manually testing this.

The coordination of certificates between the Tomcat side handling connections and the traffic router webapp side is indeed wonky. I'm working on making this much better as I address #1851 

I can always work on solving #1851 first and then applying this fix afterwards. Either way is 👍  with me.

dneuman64:
merged.  can you please cherry-pick this into the 1.7.x branch?

PR #1875, opened 2016-09-21
Change render_text to render( text => for sub postupdate in Server.pm

### LINKED ISSUE #1238: TO /update/<hostname> sends no response
When you POST valid data to the route above, Traffic Ops just closes the TCP connection, it doesn't actually send back a response. However, Traffic Ops does seem to successfully process the data that was posted.

hbeatty:
* traffic_ops/app/lib/UI/Server.pm

This should fix the 'no reply' and 500 error that was being returned when trying to clear the update column in TO.

fixes #1238

### DIFF
--- a/traffic_ops/app/lib/UI/Server.pm
+++ b/traffic_ops/app/lib/UI/Server.pm
@@ -943,14 +943,14 @@ sub postupdate {
 	my $updated   = $self->param("updated");
 	my $host_name = $self->param("host_name");
 	if ( !&is_admin($self) ) {
-		$self->render( text => "Unauthorized.", status => 401, layout => undef );
+		$self->render( text => "Forbidden", status => 403, layout => undef );
 		return;
 	}
 
 	if ( !defined($updated) ) {
 		$self->render(
 			text => "Failed request.  Must provide updated status",
-			status => 500,
+			status => 400,
 			layout => undef
 		);
 		return;
@@ -961,7 +961,7 @@ sub postupdate {
 	if ( !defined $serverid ) {
 		$self->render(
 			text => "Failed request.  Unknown server",
-			status => 500,
+			status => 404,
 			layout => undef
 		);
 		return;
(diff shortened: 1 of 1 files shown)


traffic_ops/app/lib/UI/Server.pm
```diff
@@ -943,13 +943,13 @@ sub postupdate {
 	my $updated   = $self->param("updated");
 	my $host_name = $self->param("host_name");
 	if ( !&is_admin($self) ) {
-		$self->render_text( "Unauthorized.", status => 401, layout => undef );
+		$self->render( text => "Unauthorized.", status => 401, layout => undef );
```
mitchell852: Can you change the status codes returned and see if you still get the expected results? For this one return a 403 (forbidden)

traffic_ops/app/lib/UI/Server.pm
```diff
@@ -959,8 +959,8 @@ sub postupdate {
 	# resolve server id
 	my $serverid = $self->db->resultset("Server")->search( { host_name => $host_name } )->get_column('id')->single;
 	if ( !defined $serverid ) {
-		$self->render_text(
-			"Failed request.  Unknown server",
+		$self->render(
```
mitchell852: for this one return a 404

traffic_ops/app/lib/UI/Server.pm
```diff
@@ -943,13 +943,13 @@ sub postupdate {
...
-		$self->render_text( "Unauthorized.", status => 401, layout => undef );
+		$self->render( text => "Unauthorized.", status => 401, layout => undef );
 		return;
 	}
 
 	if ( !defined($updated) ) {
-		$self->render_text(
-			"Failed request.  Must provide updated status",
+		$self->render(
+			text => "Failed request.  Must provide updated status",
```
mitchell852: for this one return 400 (Bad request)

traffic_ops/app/lib/UI/Server.pm
mitchell852: and change the text to "Forbidden"

hbeatty:
@mitchell852 All changes made as requested.

PR #1894, opened 2016-09-24
Fix tests and fixtures for pg. Update schema for tinyint to smallint …

knutsel:
…mapping

I'm pretty sure I'm doing this PR against the psql branch. Only merge it there.

### DIFF
--- a/traffic_ops/app/lib/Schema/Result/Cachegroup.pm
+++ b/traffic_ops/app/lib/Schema/Result/Cachegroup.pm
@@ -32,24 +32,22 @@ __PACKAGE__->table("cachegroup");
 
 =head2 name
 
-  data_type: 'varchar'
+  data_type: 'text'
   is_nullable: 0
-  size: 45
 
 =head2 short_name
 
-  data_type: 'varchar'
+  data_type: 'text'
   is_nullable: 0
-  size: 255
 
 =head2 latitude
 
-  data_type: 'numeric'
+  data_type: 'double precision'
   is_nullable: 1
 
 =head2 longitude
 
-  data_type: 'numeric'
+  data_type: 'double precision'
   is_nullable: 1
 
 =head2 parent_cachegroup_id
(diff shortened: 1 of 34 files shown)


traffic_ops/app/lib/Schema/Result/Cdn.pm
```diff
@@ -32,9 +32,8 @@ __PACKAGE__->table("cdn");
 
 =head2 name
 
-  data_type: 'varchar'
+  data_type: 'text'
```
mitchell852: it's my understanding that these should be 'varchar' and not 'text'. @knutsel - can you check your version of pgloader? i had this problem as well when i ran the homebrew version of pgloader. So instead of installing pgloader using homebrew, I installed version 3.3.1 directly from https://github.com/dimitri/pgloader/releases

robert-butts: Varchar is more efficient in MySQL, but PostgreSQL text has no performance or space penalty. I think it's recommended, unless you have a data constraint 

https://www.postgresql.org/docs/9.1/static/datatype-character.html

> Tip: There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs. In most situations text or character varying should be used instead.

mitchell852: I did not know that. Thanks @robert-butts

traffic_ops/app/lib/Schema/Result/Cachegroup.pm
```diff
@@ -32,24 +32,22 @@ __PACKAGE__->table("cachegroup");
...
 
 =head2 latitude
 
-  data_type: 'numeric'
+  data_type: 'double precision'
   is_nullable: 1
 
 =head2 longitude
 
-  data_type: 'numeric'
```
mitchell852: this should be 'numeric' as opposed to 'double precision' afaik because `db/admin.pl upgrade` should have run this [migration](https://github.com/Comcast/traffic_control/blob/psql/traffic_ops/app/db/migrations/20160910092026_fix_column_types.sql) which specifies the type as numeric.

PSUdaemon: I concur, this should be reverted back.

mitchell852: I created a new PR that sets appropriate columns to numeric instead of double precision - https://github.com/Comcast/traffic_control/pull/1907

PR #1899, opened 2016-09-26
Fix TM2 DSStats to a readonly interface

### LINKED ISSUE #1867: Fix Traffic Monitor 2.0 DSStatsThreadsafe.Get() to not modify
Right now, if DSStatsThreadsafe uses a RLock, the app eventually panics from multiple concurrent map writes. Thus, either Get() itself is mutating data, or some data (a member map) is being passed without copying, and the caller is mutating.

Either way, it should be fixed, and RLock used, which is faster than Lock.

We should consider changing DSStatsThreadsafe (and the other Threadsafe types) to return interfaces with only accessor functions, so it's impossible for callers to mutate. I would've considered that YAGNI, except we do need it, it is happening.

robert-butts:
Merge issue #1887 before this. They'll have merge conflicts, and it'll be easier to resolve this than that.

Fixes #1867


dneuman64:
@robert-butts this is ready for you to fix merge issues

robert-butts:
Fixed

PR #1921, opened 2016-09-28
TO: updates dockerfile for centos7

dangogh:


mitchell852:
I'll defer to @dewrich on this one...

PR #1926, opened 2016-09-28
Change TO client to not log

### LINKED ISSUE #1925: Change Traffic Ops Client to not log
In general, libraries shouldn't log. In specific, it's filling Traffic Monitor 2.0 logs with a different format, and which can't be turned off or down.

The Traffic Ops Client should be changed to not log, and to return all logged data (e.g. cache hit/miss status), so users of the library can decide whether to log.

robert-butts:
Changes the Traffic Ops Client to not log, and to return all formerly
logged information so users can decide whether to log, and in their
own format.

Fixes #1925

### DIFF
--- a/traffic_monitor/experimental/traffic_monitor/trafficopswrapper/trafficopswrapper.go
+++ b/traffic_monitor/experimental/traffic_monitor/trafficopswrapper/trafficopswrapper.go
@@ -27,7 +27,7 @@ func (s TrafficOpsSessionThreadsafe) CRConfigRaw(cdn string) ([]byte, error) {
 	if s.session == nil || *s.session == nil {
 		return nil, fmt.Errorf("nil session")
 	}
-	b, e := (*s.session).CRConfigRaw(cdn)
+	b, _, e := (*s.session).GetCRConfig(cdn)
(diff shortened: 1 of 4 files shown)


dangogh:
this changes the signature on several published functions.    Where are those being used?

robert-butts:
I searched, and didn't find any but TM2.0. Traffic Stats builds, and the tests pass. Did I miss any?

dangogh:
well,  it's not just available internally..  could an external user have created something that uses it?   I don't know of any offhand, but there's really no good way to know for sure..   Just trying to make sure it's considered...

robert-butts:
Yeah, you're right. It seems unlikely, but since we have a release with it, we shouldn't break compatibility.

I'll change those signatures back, and add new functions with the status data.

Though, it seems silly to have two functions long-term, so I'll add a GoDoc 'Deprecated' comment, and we can remove them in 3.0.

dangogh:
if there's a straightforward way to make the old one a wrapper around the new one, that could work too..   But I'd be ok with it either way..   Assigning to myself since I've already looked at it a bit..

robert-butts:
Fixed.

PR #1982, opened 2016-10-06
Traffic Stats now uses its config file for InfluxDb Connections

### LINKED ISSUE #1668: Enhance Traffic Stats to write to multiple InfluxDb servers.
Currently, Traffic Stats uses Traffic Ops to get an online InfluxDb server from the database and write to it.  The assumption is that Traffic Stats is either writing to a single influxDb instance or one node in a cluster.  Since the latest versions of InfluxDb do not support clustering for free, Traffic Stats should be enhanced such that Traffic Stats attempts to write each point to all of the InfluxDb servers listed as ONLINE in Traffic Ops.  If Traffic Stats cannot write to one of the InfluxDb servers listed, it should log a warning, if Traffic Stats cannot write to any of the InfluxDb servers listed, it should log an error.  

This feature will serve as an initial workaround for clustering support with newer versions of InfluxDb, a longer term solution still needs to be engineered.

dneuman64:
Traffic Stats now uses its config file to determine which influxdb urls to use.  This allows a user to run a single instance of influxdb, an enterprise clustered solution, or a HA solution using influxdb-relay (https://github.com/influxdata/influxdb-relay).  

This version of Traffic Stats requires Influxdb version 1.0.0 or greater.

This fixes #1668

### DIFF
--- a/traffic_stats/influxdb_tools/sync_ts_databases.go
+++ b/traffic_stats/influxdb_tools/sync_ts_databases.go
@@ -439,7 +439,7 @@ func getDeliveryServiceStats(res []influx.Result) map[string]deliveryServiceStat
 
 func getDailyStats(res []influx.Result) map[string]dailyStats {
 	response := make(map[string]dailyStats)
-	if res != nil && len(res[0].Series) > 0 {
+	if len(res) > 0 && len(res[0].Series) > 0 {
 		for _, row := range res[0].Series {
 			for _, record := range row.Values {
 				data := new(dailyStats)
--- a/traffic_stats/influxdb_tools/sync_ts_databases.go
+++ b/traffic_stats/influxdb_tools/sync_ts_databases.go
@@ -439,7 +439,7 @@ func getDeliveryServiceStats(res []influx.Result) map[string]deliveryServiceStat
 
 func getDailyStats(res []influx.Result) map[string]dailyStats {
 	response := make(map[string]dailyStats)
-	if len(res[0].Series) > 0 {
+	if res != nil && len(res[0].Series) > 0 {
 		for _, row := range res[0].Series {
 			for _, record := range row.Values {
 				data := new(dailyStats)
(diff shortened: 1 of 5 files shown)


traffic_stats/traffic_stats.cfg
```diff
@@ -13,5 +13,6 @@
 	"cacheRetentionPolicy": "daily",
 	"dsRetentionPolicy": "daily",
 	"dailySummaryRetentionPolicy": "indefinite",
-	"influxProtocol": "http"
+	"influxProtocol": "http",
```
PSUdaemon: Do you need `influxProtocol` anymore?

traffic_stats/influxdb_tools/sync_ts_databases.go
```diff
@@ -439,7 +439,7 @@ func getDeliveryServiceStats(res []influx.Result) map[string]deliveryServiceStat
 
 func getDailyStats(res []influx.Result) map[string]dailyStats {
 	response := make(map[string]dailyStats)
-	if len(res[0].Series) > 0 {
+	if res != nil && len(res[0].Series) > 0 {
```
robert-butts: `if len(res) == 0` would be safer here. I don't know if `res` will ever be constructed empty, but if it is, `res` could be non-nil and `res[0]` would still panic

traffic_stats/traffic_stats.cfg
dneuman64: I am currently using the to determine if we connect via udp or http,  see line 757.  The alternative would be to parse the Urls provided...

traffic_stats/influxdb_tools/sync_ts_databases.go
robert-butts: `if len(res) > 0` would be safer here. I don't know if res will ever be constructed empty, but if it is, res could be non-nil and `res[0]` would still panic

traffic_stats/traffic_stats.cfg
PSUdaemon: Yeah, since it's in the URL, I think we should utilize that. Otherwise they could conflict.

PR #1993, opened 2016-10-07
use env defaults if unset for docker build clone

dangogh:
allows `docker-compose up` with no env variables to build from default repo/branch


dangogh:
Hold off on this -- found an issue with it..

dangogh:
Now it's ready.....

Issue #61, opened 2015-04-17
Routing name/entry point in CRConfig


elsloo:
A default routing name/entry point should be added as a profile parameter, and this parameter should be used by default as the value of the "routing name" or "entry point" on the delivery service in the CRConfig.  This value should be overridable on a per-delivery service level, which means that the delivery service UIs should be enhanced to provide this capability and to display the default value.

elsloo:
This is related to and/or required by issue #53.

limited:
https://issues.apache.org/jira/browse/TC-287

PR #2002, opened 2016-10-11
Changed ccr. to tr. for Example Delivery Service URLs

sandmanahoy:


mitchell852:
These "example" URLs are the actual URLs used as the entry point into the CDN for the delivery service. If *DNS*, the URL starts with 'edge', else the URL starts with 'ccr'. 

However, you have a good point, others may decide to use "tr" (traffic router) instead of 'ccr'. therefore, these values should really be data-driven in my opinion. For example, you could add a parameter to the global profile in the seeds.sql file and then you could query for the value of that parameter. Here's an example in the codebase of where that was done - https://github.com/Comcast/traffic_control/blob/master/traffic_ops/app/lib/API/Job.pm#L209

knutsel:
yea i think there's a param somewhere that decides if you're using ccr, tr, or something else..... that param should be used. (could've sworn that we added that code, but i guess i'm wrong.). 

@smalenfant  - can't find that param, but I know you're running with tr. ... do you remember how to set the ccr/tr prefix?

mtorluemke:
it's this property in applicationContext.xml:

<property name="httpRoutingName" value="$[http.routing.name:tr]" />

https://github.com/Comcast/traffic_control/blob/master/traffic_router/core/src/main/webapp/WEB-INF/applicationContext.xml

trevorackerman:
Mark is correct.
You would add something like the following line in the file /opt/traffic_router/conf/http.properties
`http.routing.name=foo`

elsloo:
@mitchell852 there was a PR and issue for configurable routing names, however, the PR had merge conflicts and we didn't have the resources to finish the work. PR #201 and issue #61.

sandmanahoy:
@mtorluemke that agrees with my understanding as well, the parameter is configurable in http.properties in TR but that file is not managed by TO. The parameter defaults to TR.

traffic_ops/app/lib/UI/DeliveryService.pm
```diff
@@ -147,9 +147,9 @@ sub get_example_urls {
 					? $example_urls[ $re->{set_number} ]
 					: "";
 				if ( $re->{set_number} == 0 ) {
-					$example_urls[ $re->{set_number} ] = $scheme . '://ccr.' . $host . "." . $cdn_domain . $p;
+					$example_urls[ $re->{set_number} ] = $scheme . '://tr.' . $host . "." . $cdn_domain . $p;
```
mitchell852: this won't work. this hardcodes it in traffic ops to 'tr' (instead of ccr) and from what i heard it can really be anything that you want it to be as defined in https://github.com/Comcast/traffic_control/blob/master/traffic_router/core/src/main/webapp/WEB-INF/applicationContext.xml in the http.routing.name property. 

see what i mean? line 150 & 152 need to be dynamic, not hard-coded. if it were me, i'd create a global parameter using a database migration file and call the parameter http.routing.name and set the default value to 'tr'. then in the code, i'd fetch the value of http.routing.name and use that on line 150 & 152. 

migration files are found here: https://github.com/Comcast/traffic_control/tree/master/traffic_ops/app/db/migrations

elsloo: As a half step before #61 is implemented (it's still open)? Personally I'm +1 on this as a quick change to get the UI in line with our defaults; this is better than what is there currently.

mitchell852: the only problem is that these "example" urls are exposed thru the traffic ops API which means they show up in the traffic portal - https://github.com/Comcast/traffic_control/blob/master/traffic_ops/app/lib/API/DeliveryService.pm#L80 

so if you have http.routing.name=foo in applicationContext.xml, the example urls will still show as tr.ds-name.example.domain.com rather than the correct url of foo.ds-name.example.domain.com

elsloo: Good point about the portal. The main issue is we use the portal, many folks probably don't. So, if we change the default, our portal customers will see the "tr," and we have no means to configure it, which will lead to confusion. I agree that this half-step is probably warranted mostly due to the portal dependency.

mitchell852:
@sandmanahoy - per [CONTRIBUTING.md](https://github.com/Comcast/traffic_control/blob/master/CONTRIBUTING.md) you'll need to sign a CLA and submit to @knutsel at JAN_VANDOORN@COMCAST.COM before this PR can be merged.

Also, I've created an issue for this PR - https://github.com/Comcast/traffic_control/issues/2025

knutsel:
@mitchell852 : @sandmanahoy is covered under the Cisco CLA. Ok to merge.

PR #2022, opened 2016-10-14
Fix TM2 to return 503 until all caches are polled

### LINKED ISSUE #2017: Fix Traffic Monitor 2.0 To Not Serve Until All Caches Are Polled
TM2 needs to not serve endpoints until all caches have been polled twice, and we have bandwidths.

This requires a reasonably short fetch timeout, and we need to flag timed out requests as having been initially polled, so we don't wait forever for down caches.

We also need to determine the existing TM1 behavior to replicate. Does it not serve? Does it return a 404? Some sentinel message?

robert-butts:
Fixes Traffic Monitor 2.0 to return 503 Service Unavailable to all
requests, until all caches have either been polled twice and have
a bandwidth set, or have failed to respond properly.

This is necessary, because otherwise, interdependent components such
as Traffic Router could request and get a partial result, with wrong
availability, bandwidth, and other numbers.

Fixes #2017

### DIFF
--- a/traffic_monitor/experimental/traffic_monitor/manager/datarequest.go
+++ b/traffic_monitor/experimental/traffic_monitor/manager/datarequest.go
@@ -419,23 +419,30 @@ func DataRequest(
 	toData todata.TODataThreadsafe,
 	localCacheStatus CacheAvailableStatusThreadsafe,
 	lastStats LastStatsThreadsafe,
+	unpolledCaches UnpolledCachesThreadsafe,
 ) (body []byte, responseCode int) {
+
 	// handleErr takes an error, and the request type it came from, and logs. It is ok to call with a nil error, in which case this is a no-op.
-	handleErr := func(err error, requestType http_server.Type) {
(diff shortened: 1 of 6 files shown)


traffic_monitor/experimental/traffic_monitor/manager/datarequest.go
```diff
@@ -419,83 +419,90 @@ func DataRequest(
...
 			return body, http.StatusOK
 		}
-		handleErr(err, requestType)
+		handleErr(err)
 		return nil, http.StatusInternalServerError
 	}
 
+	if unpolledCaches.Any() {
+		handleErr(fmt.Errorf("service still starting, some caches unpolled"))
+		return []byte("Service Unavailable"), http.StatusServiceUnavailable
```
robert-butts: I was tempted to make this `return []byte("We don't want any more visitors, well-wishers or distant relations!"), http.StatusServiceUnavailable` >_>

